PR #84, opened 2012-09-07
Editing while reading.

jean:
Just copy-editing while reading. 
- Use spaces, not tabs.
- Avoid lines longer than around 80 characters, and break lines on phrases (just better for diffing documentation).
- Specify console highlight where appropriate.
- Fix an extra comma here and there. 
- Fix a typo or two (e.g. "Philosphy").

### DIFF
--- a/tablib/core.py
+++ b/tablib/core.py
@@ -134,7 +134,7 @@ class Dataset(object):
 
 
     You can also set rows and headers upon instantiation. This is useful if dealing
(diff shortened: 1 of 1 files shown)


jean:
@kennethreitz it looks like development has picked up again and the PR is still merge-able .. would you like to take another look at it?

jean:
On second thoughts, I see it's targeting a now non-existent `develop` branch. I'll see what it looks like against master ..

PR #163, opened 2014-08-27
LaTeX table export format

mloesch:
Lets you export your data as booktabs-styled LaTeX table.

Note that the styling uses some reasonable defaults (IMHO) but cannot be perfect. It is basically intended to give you a starting point and save you from writing the cumbersome LaTeX table format yourself.

Also note that multi-line fields currently are not supported (export works but the resulting LaTeX code won't compile).

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -29,3 +29,4 @@ Patches and Suggestions
 - Alex Gaynor
 - James Douglass
 - Tommy Anthony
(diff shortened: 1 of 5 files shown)


kennethreitz:
merged!

kennethreitz:
(bet you didn't expect that :P)

PR #164, opened 2014-08-30
Creating subset for Datasets + some minor fixes

rabinnankhwa:
A new function "subset" was added which can create a separate dataset containing selected rows and columns present in the current dataset.
A bug in __getslice__ method was fixed. Also, built-in get method for dictionary was used instead of try-except structure.


kennethreitz:
:sparkles: :cake: :sparkles:

Sorry for taking so long!

PR #182, opened 2015-02-04
Adding ability to unique all rows in a dataset.

cherepski:
This commit adds in functionality to unique all rows within a dataset while maintaining the original dataset ordering.


kennethreitz:
Thanks! I'm going to rename this `remove_duplicates`.

Issue #167, opened 2014-09-02
Update openpyxl


rattrayalex:
It looks like tablib is still using openpyxl from three years ago. The [open source package](https://pypi.python.org/pypi/openpyxl) is actively maintained, receiving several major updates even in the last 6 months. 

The version used by tablib cannot even open xlsx documents. 
Eg; 

``` python
with open('myfile.xlsx', 'rb') as f:
    databook = tablib.import_book(f.read())
```

results in the following traceback: 

```
TypeError                                 Traceback (most recent call last)
<ipython-input-5-e85247aaa4e8> in <module>()
      1 with open('_data/members_old.xlsx', 'rb') as f:
----> 2     ds = tablib.import_book(f.read())
      3

/usr/local/lib/python2.7/site-packages/tablib/core.pyc in import_book(stream)
   1016 def import_book(stream):
   1017     """Return dataset of given stream."""
-> 1018     (format, stream) = detect(stream)
   1019
   1020     try:

/usr/local/lib/python2.7/site-packages/tablib/core.pyc in detect(stream)
    994     for fmt in formats.available:
    995         try:
--> 996             if fmt.detect(stream):
    997                 return (fmt, stream)
    998         except AttributeError:

/usr/local/lib/python2.7/site-packages/tablib/formats/_xlsx.pyc in detect(stream)
     29     """Returns True if given stream is a readable excel file."""
     30     try:
---> 31         openpyxl.reader.excel.load_workbook(stream)
     32         return True
     33     except openpyxl.shared.exc.InvalidFileException:

/usr/local/lib/python2.7/site-packages/tablib/packages/openpyxl/reader/excel.pyc in load_workbook(filename, use_iterators)
     66
     67     try:
---> 68         archive = ZipFile(filename, 'r', ZIP_DEFLATED)
     69     except (BadZipfile, RuntimeError, IOError, ValueError):
     70         raise InvalidFileException()

/usr/local/Cellar/python/2.7.8/Frameworks/Python.framework/Versions/2.7/lib/python2.7/zipfile.pyc in __init__(self, file, mode, compression, allowZip64)
    754             modeDict = {'r' : 'rb', 'w': 'wb', 'a' : 'r+b'}
    755             try:
--> 756                 self.fp = open(file, modeDict[mode])
    757             except IOError:
    758                 if mode == 'a':

TypeError: file() argument 1 must be encoded string without NULL bytes, not str
```

claudep:
[...]

PR #185, opened 2015-02-22
De-vendorized xlrd

jsma:
Bundled version was ancient and incompatible with modern versions of xlsx


jsma:
Refs #167

jsma:
Was this closed because the problems mentioned are solved now or for other reasons (poor quality PR, etc)? Thanks for creating this btw!

PR #186, opened 2015-03-02
Added a mechanism to avoid datetime.datetime issues when serializing dat...

hdzierz:
Hello,

I was wondering whether we could add the lines of the pull request to tablib. It p[revents an error message when serializing datetime.datetime objects.

Up to you.

Cheers

H


kennethreitz:
:sparkles: :cake: :sparkles:

PR #189, opened 2015-03-28
Fixes Row slicing. Fixes #184.

### LINKED ISSUE #184: Error in Row.__getslice__ method
If trying to do `row[1:5]` (`row` being of type `Row`) the error `TypeError: list indices must be integers, not tuple` is thrown.

The cause seems to be the `__getslice__` method:

```python
    def __getslice__(self, i, j):
        return self._row[i,j]
```
(class `Row` in core.py:47)

I think replacing the comma by a colon should do the trick:

```python
    def __getslice__(self, i, j):
        return self._row[i:j]
```

PS Thanks for building tablib! It saves a ton of work.

tsroten:
Issue #184 reports a bug in the `Row.__getslice__` method. This commit fixes that bug and includes a simple test to verify `Row` slicing works.

### DIFF
--- a/tablib/core.py
+++ b/tablib/core.py
@@ -45,7 +45,7 @@ class Row(object):
         return repr(self._row)
 
     def __getslice__(self, i, j):
-        return self._row[i,j]
+        return self._row[i:j]
 
     def __getitem__(self, i):
         return self._row[i]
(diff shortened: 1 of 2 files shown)


iurisilvio:
Thanks!

PR #193, opened 2015-04-08
Formats .tsv and .html are implemented by now

jhermann:
Removed mentioning of "wanted" formats that exist.


kennethreitz:
Thanks!

PR #195, opened 2015-05-28
Removed the vendorized packages

tommyanthony:
I've removed the vendorized versions of xlwt, xlrd, openpyxl, ominjson, unicodecsv, and pyyaml and replaced them by installing them in setup.py.

If there's a reason this hasn't been done already, I'd love to know and see if there's anything we can do to update the vendorized versions to newer releases.

Edit: Ok, I now see it fails on python 3.2.  Not sure how important that is, but if that's why this hasn't been done this can be closed I guess.


iurisilvio:
I want something like this too, it should be done, but don't know the right way to do it. That is the reason it is not done. I don't want to enforce the installation of all these packages.

I'll maintain it open and will try to make it happen soon.

The Python 3.2 support definitely is not a good reason, we can drop the support for this version.

tommyanthony:
Having dependencies installed via pip is pretty normal, see pyspider's setup.py: https://github.com/binux/pyspider/blob/master/setup.py#L56

Also, I noticed the last time the PyPI version was updated was a year ago, do you know how to update the version installed by pip?

iurisilvio:
I agree, but I'm looking for a solution with `extras`. I'm planning to release a `1.0` soon, but no fixed dates yet. This is one big change I want to do.

tommyanthony:
I updated the pull request, maybe you'll like this one more.  I removed the requirement of installing omnijson, and made it only install unicodecsv if using python 2.  I also removed python 3.2 from .travis.yml.  

If tablib can import omnijson it will still use it, otherwise it will default to the standard json library.

claudep:
Looks like this PR is in good shape, doesn't it? Iuri, any remaining issue?

kennethreitz:
The vendored packages will stay for.. they work, and they allow you to use this package without installing 200 dependencies.

There are many Python 3 ports within this codebase that I did myself, because these projects were abandoned at the time. Hopefully this is no longer the case, but this will require a closer investigation.

claudep:
200 dependencies? Aren't we talking about ~6 here?
A closer investigation? Aren't you trusting your own test suite?
You are the maintainer, you decide, I just find a bit sad to just put away contributed work with almost no rationale.

kennethreitz:
@claudep don't worry, I'll likely accept this soon. Your work is very appreciated. Just needs some time to brew.

I'll keep the PR open for peace of mind :)

claudep:
To be clear, it's not my work at all (I probably won't dare to say that for my own work :-) ).
Sorry if I misunderstood that closing the PR was simply ditching this work. If it's not the case, it's good news!

kennethreitz:
@claudep can you review the failing tests?

kennethreitz:
er, @tommyanthony can you review the failing tests?

claudep:
My two cents: for the first failure:
```
======================================================================
ERROR: test_book_export_no_exceptions (__main__.TablibTestCase)
Test that various exports don't error out.
----------------------------------------------------------------------
  File "test_tablib.py", line 414, in test_book_export_no_exceptions
    book.xlsx
  File "/home/travis/build/kennethreitz/tablib/tablib/formats/_xlsx.py", line 54, in export_book
    wb.worksheets = []
AttributeError: can't set attribute
```
In recent openpyxl, `wb.worksheets` is a property (iterating over `wb._sheets`). To clear sheets, you can either access the "private" _sheets attribute, or delete existing sheets in a loop with public API: `for sheet in  wb.worksheets: wb.remove_sheet(sheet)`.

claudep:
I tried to concentrate on openpyxl only (and specific test failures) in #217. If accepted, we could then do the same for other packages.

PR #204, opened 2015-10-07
Replace tabs with whitespaces

dallagi:
According to pep8, spaces should be used to indent code [0].
The file formats/_html , unlike all the other files, is indented with tabs.
This pull request replaces tabs with spaces in that file.

[0] https://www.python.org/dev/peps/pep-0008/#tabs-or-spaces


kennethreitz:
When this code was written, PEP 8 stated that either spaces or tabs are fine, as long as they are used consistently :)

Thanks!

PR #208, opened 2015-11-09
Fix XLSX import

stclair:
Calling import_set on an XLSX file was throwing a TypeError from
Openpyxl. Openpyxl Reader load_workbook requires a file-like object as the first
argument. This commit fixes the error by passing in a file-like object
instead of a string.

### DIFF
--- a/tablib/formats/_xlsx.py
+++ b/tablib/formats/_xlsx.py
@@ -69,7 +69,7 @@ def import_set(dset, in_stream, headers=True):
 
     dset.wipe()
 
-    xls_book = openpyxl.reader.excel.load_workbook(in_stream)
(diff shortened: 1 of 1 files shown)


stclair:
If this looks alright, I'll submit another PR to update the documentation here: https://github.com/stclair/tablib/blob/develop/tablib/core.py#L456

If we can create a release including this PR, and I can update https://github.com/django-import-export/django-import-export to include XLSX import support.

Thank you,
Wes

stclair:
@kennethreitz I noticed this PR had a temp commit that shouldn't have been there.  I updated it.  Would you consider merging this in?  Let me know if you would like tests added for this, or anything else.

kennethreitz:
:sparkles: :cake: :sparkles:

Issue #53, opened 2012-02-07
xlsx output has freeze panes on wrong axis


slestak:
I have not fully nailed this down yet.  I definitely have sample files created from datasets with say 11 cols x 300 rows and the freeze pane was running vertical at the next to last column.

I looked at https://github.com/kennethreitz/tablib/blob/develop/tablib/formats/_xlsx.py#L77 and counter that it could be simplified 
from:

ws.freeze_panes = '%s%s' % (col_idx, row_number)

to:

ws.freeze_panes = 'A2'

PR #216, opened 2016-02-08
Error: frzn_col_idx not defined

cmhofer:
As in #53 already solved. Freeze panes set to A2 again.

See commit 954bbdccf32b42bb2871a6dc851eec476fee803a where it was solved but lost in merge 8debeb26aca7cc9ef427c70ee9d577eb169f9560


kennethreitz:
Ah, I was wondering about that. Thanks!

PR #219, opened 2016-02-16
Fix export only formats

timofurrer:
Formats like LaTeX could have never been exported because
`setattr(cls, set_%s % fmt.title, fmt.import_set)` always failed
for export-only formats and with that the exception was caught in the
outer try/except and the format tuple was set to (None, None) with
`cls._formats[fmt.title] = (None, None)`


kennethreitz:
Thank you very much!

PR #228, opened 2016-02-25
Fixed textual representation for Dataset with no headers

tomchristie:
If a `Dataset` does not have any headers, then it is not printable:

    >>> data = tablib.Dataset(['a', 1], ['b', 2], ['c', 3])
    >>> print data
      File "<stdin>", line 1, in <module>
      File ".../tablib/core.py", line 242, in __str__
        return self.__unicode__()
      File ".../tablib/core.py", line 226, in __unicode__
        result.append([unicode(h) for h in self.__headers])
    TypeError: 'NoneType' object is not iterable

This change resolves this, using a string representation of the dataset, without headers:

    >>> data = tablib.Dataset(['a', 1], ['b', 2], ['c', 3])
    >>> print data
    a|1
    b|2
    c|3

### DIFF
--- a/tablib/core.py
+++ b/tablib/core.py
@@ -223,7 +223,8 @@ class Dataset(object):
         result = []
 
         # Add unicode representation of headers.
-        result.append([unicode(h) for h in self.__headers])
+        if self.__headers:
+            result.append([unicode(h) for h in self.__headers])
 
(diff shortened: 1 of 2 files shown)


kennethreitz:
ooh, i like! 

Thanks :)

PR #232, opened 2016-03-18
[docs] Update variable name in tuto

chimeno:
The tutorial has been using the 'data' variable, but in this case it's using 'd'.
This change that.


kennethreitz:
✨ 🍰 ✨

PR #233, opened 2016-03-18
Add HTML format to the book_export test and fix the format to work properly

ScorpionResponse:
The export_book method in the _html.py format file was not being called by tests.  Adding it causes this error:

```
ERROR: test_book_export_no_exceptions (__main__.TablibTestCase)
Test that various exports don't error out.
----------------------------------------------------------------------
  File "test_tablib.py", line 395, in test_book_export_no_exceptions
    book.html
  File "/home/vagrant/development/tablib/tablib/formats/_html.py", line 63, in export_book
    stream.write('<%s>%s</%s>\n' % (BOOK_ENDINGS, title, BOOK_ENDINGS))
TypeError: 'str' does not support the buffer interface
```

This PR also fixes that error largely by copying the way that export_set method does.

### DIFF
--- a/tablib/formats/_html.py
+++ b/tablib/formats/_html.py
@@ -31,7 +31,7 @@ def export_set(dataset):
     page.table.open()
 
     if dataset.headers is not None:
-        new_header = [item if item is not None else '' for item in dataset.headers] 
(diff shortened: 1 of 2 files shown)


kennethreitz:
Thanks!

PR #234, opened 2016-03-23
if the object is a decimal, return the string representation of it.

BrianPainter:
in addition to dates, also handle python decimals (convert them to a string).


kennethreitz:
This seems reasonable.

BrianPainter:
anything else I need to do to help this get in master?

PR #244, opened 2016-07-02
Rewrite ODS support based on loxun XMLWriter module

bdauvergne:
It uses constant memory and is a lot faster than odf and odf3 packages as the document is not built in memory prior to serialization. OpenDocument is a simple format that should not need many thousand lines of code and gigabytes of memory to export a simple table of tens of thousand of lines.

A temporary file is needed as zipfile does not support streaming directly into it, if it's a problem I can do it in memory with BytesIO augmenting a little bit the memory consumption.

With the current implementation it's nearly impossible to export a 100 000 lines table to ODS in a constrained memory environment (VM with 1 Gb of memory).


chfw:
@bdauvergne , just out of curiosity, could I find the ods writer lib(Copyright (C) 2005-2016  Entr'ouvert) on pypi or github?

bdauvergne:
This code is new, I produced it on my employer (Entr'ouvert) time, it's freely inspired by this package (http://git.entrouvert.org/wcs.git/tree/wcs/qommon/ods.py) also from Entr'ouvert which use ElementTree and so do not have bounded memory consumption for this you need a streaming XmlWriter like API.

chfw:
Thanks for your reply. 

I planned to copy your code to produce a specialised ods writer for pyexcel, as pyexcel-odsw. As you mentioned in this PR, odfpy and ezodf does not use constant memory in writing an ods. I hope you will be OK with my copying.

For your information, messy-tables had a better performing ods reader and it inspired [pyexcel-odsr](https://github.com/pyexcel/pyexcel-odsr). So your code is the missing puzzle to complete ods story.

bdauvergne:
No problem, just keep the copyright.

tablib/packages/ods.py
```diff
@@ -0,0 +1,150 @@
...
+        assert self.status == self.INSHEET
+        self.status = self.INROW
+        self.xmlwriter.startTag('table:table-row', attribs)
+
+    def end_row(self):
+        assert self.status == self.INROW
+        self.status = self.INSHEET
+        self.xmlwriter.endTag()
+
+    def add_cell(self, content, hint=None):
```
chfw: Just an observation here. It is not a bug or anything.

`add_cell` does not support other cell data types, such as: int, float but unicode string.

bdauvergne: Yep, small amerliorations are still possible, I would do it if i had information from the maintainer that a possible integration is possible soon.

chfw: Sent an invitation to you.

frallain:
@bdauvergne Why not add loxun in the `requirements.txt` as it is available on pypi at https://pypi.org/project/loxun/ instead of copy pasting the whole file in the `tablib` project?

bdauvergne:
Just thought it was the tablib way, it contains (contained?) so much external dependencies, I did not know they were all not packaged on pypi.

PR #245, opened 2016-07-05
HTML input: formats.html.import_set

klothe:
formats.html.import_set imports a Dataset from an HTML table.

It assumes there is exactly one table in the HTML document. (An import_book function would be more appropriate to handle multiple tables, but that doesn't seem very useful.) Headers are taken from <th> elements if they exist.

BeautifulSoup (bs4) is used for HTML parsing, assuming it's OK to add that dependency. I didn't add a copy of bs4 to the "packages" directory since I'm not sure if that is necessary, or why it was done with other dependencies. Alternately, it could be added as a requirement in setup.py. Which way would be best?

### DIFF
--- a/tablib/formats/_html.py
+++ b/tablib/formats/_html.py
@@ -11,6 +11,7 @@ if sys.version_info[0] > 2:
 else:
     from cStringIO import StringIO
     from tablib.packages import markup
+import bs4
 
 import tablib
 from tablib.compat import unicode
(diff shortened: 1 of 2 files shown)


timofurrer:
Please resolve the merge conflicts. Thanks :tada:

claudep:
I worked on similar functionality, but trying without any external dependency: #555.

PR #269, opened 2017-01-10
Fixed classifiers in setup.py

kammala:
moved classifiers from tuple to list(this allow to use `setup.py upload` command in python >= 3.5)
added python 3.5 and 3.6 to classifiers list


iurisilvio:
Thanks!

PR #276, opened 2017-02-18
correct example

yarko:
map() is a function in python2, and iterator in python3+;

In any case map is inefficient compared to either comprehensions (most efficient), 
   or simple loops (close second).
SInce in this case, data.append() returns nothing, use a simple loop.
It is clearer, more efficient, and works with both python2 and python3


yarko:
...btw, just for the heck of it, I ran this on cPython 3.7a (so far, seems to work);

yarko:
sorry - just realized I should have made this against "develop" branch;  let me know if you need me to do anything...

Issue #273, opened 2017-01-25
Unbundle all dependencies. 


kennethreitz:
It's time.

iurisilvio:
I agree! But why you closed the issue?

kennethreitz:
@morleyb might take a crack at this!

hermanocabral:
please!

kennethreitz:
@morleyb need any help getting started?

jayvdb:
De-vendoring dbfpy is going to be a bit more effort than the others.

The dbfpy package on PyPi is 2.3.1 , and the bundled version here is from an older release, 2.2.5, and it is slightly patched, *and* the tests here fail when using the latest PyPI package - the struct's are different.

The errors can be encountered with https://github.com/jayvdb/tablib/commit/95cfef3ae88d50848776165fec345f3d0a29e869, and viewed at https://travis-ci.org/jayvdb/tablib/jobs/505156956

The adaptions are not visible in git, but there is a bit of history at https://github.com/kennethreitz/tablib/pull/153 , and a bit more at https://github.com/phargogh/dbfpy3/commits/master

@phargogh might be able to help.

The diff which makes the tablib tests pass is

```patch
--- dbfpy-2.3.1/dbfpy/fields.py.orig    2019-03-12 16:16:12.434154549 +0700
+++ dbfpy-2.3.1/dbfpy/fields.py 2019-03-12 16:33:19.785761986 +0700
@@ -150,7 +150,7 @@
         return (
             _name +
             self.typeCode +
-            struct.pack("<L", self.start) +
+            chr(0) * 4 +
             chr(self.length) +
             chr(self.decimalCount) +
             chr(0) * 14
```

But I suspect that correct answer will be a more complicated mixture of the two codebases, as both have had bug fixes, and dbf comes in several flavours.

A few days ago I found another version of the codebase on GitHub, and asked if they are interested in releasing a new package at https://github.com/estelendur/dbfpy/issues/1 , and they are interested.

jayvdb:
Another approach is to switch to using a different library, such as pypi packages dbfread or dbf.

PR #277, opened 2017-02-20
Replaced vendored xlrd/xlwt by dependencies

claudep:
Refs #273.


vToMy:
This is also an important PR, especially since the xlrd package does not even compile with python3, which is an issue when trying to distribute with pyinstaller.

PR #279, opened 2017-02-21
Replaced vendored omnijson by the standard lib version

claudep:
Refs #273.


setup.py
```diff
@@ -14,16 +14,6 @@
...
     sys.exit()
 
-if sys.argv[-1] == 'speedups':
-    try:
-        __import__('pip')
-    except ImportError:
-        print('Pip required.')
-        sys.exit(1)
-
-    os.system('pip install ujson')
```
iurisilvio: I think it removed a feature (speed). We can try to import ujson if it is available.

```
try:
    import ujson as json
except ImportError:
    import json

claudep: When reading the numbers on https://pypi.python.org/pypi/ujson, it seems own Python json is as fast or even faster, or did I misread them?

iurisilvio: `ujson` is faster. The numbers in their benchmarks are `number of operations`. Higher is better.

claudep: Just found http://jyotiska.github.io/blog/posts/json_vs_simplejson_vs_ujson.html
Then let's keep the ujson import, I'll update the patch.

claudep: Do you also want to keep the "speedups" part in setup.py?

iurisilvio:
Yes, maintain the speedups.

Em 24 de fev de 2017 07:14, "Claude Paroz" <notifications@github.com>
escreveu:

claudep:
Should be good to go now.

iurisilvio:
Great! Thanks!

PR #294, opened 2017-08-07
Switched format and stream arguments in import_book call to Databook.load

JorikBooij:
The arguments (stream and format) in the Databook.load call from import_book are in the wrong order. This causes Databook.load to always throw an UnsupportedFormat exception.

Could also be fixed by switching the arguments on Databook.load, but that might break more existing code.


timofurrer:
Already fixed with: https://github.com/kennethreitz/tablib/pull/315

PR #307, opened 2017-09-01
Make pandas optional

### LINKED ISSUE #306: Pandas support
Can we make a tablib[pandas] optional install? with 0.12.0 there is now a pandas dependency that is a huge dependency to add for everyone by default.

audiolion:
Resolves #306 

Not sure if this is the best way to notify the user about not having pandas installed when using the df package, one suggestion was to use a system like:

https://github.com/habnabit/passacre/blob/master/passacre/features.py
https://github.com/habnabit/passacre/blob/master/passacre/config.py#L103

which is pretty cool but I didn't know about adding their license in with this project and making so many modifications.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,5 +6,5 @@ python:
   - 3.5
   - 3.6
 install:
-  - python setup.py install
+  - pip install -r requirements.txt
 script: python test_tablib.py
--- a/setup.py
+++ b/setup.py
@@ -48,7 +48,6 @@ install = [
     'xlrd',
     'xlwt',
     'pyyaml',
-    'pandas'
 ]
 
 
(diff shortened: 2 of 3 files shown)


kennethreitz:
Also gotta update the tests!

kennethreitz:
Actually, that's my fault.

audiolion:
yeah my bad I didn't realize 3.3 support was there

audiolion:
pretty close to EOL for 3.3 https://www.python.org/dev/peps/pep-0398/#x-end-of-life

kennethreitz:
ugh, i'll push a release now

kennethreitz:
v0.12.1 released!

Issue #252, opened 2016-09-06
xlsx column format


farzeni:
Hello, 

thanks for your good work.

I have an issue exporting data in xlsx format. I have a data table with different type of columns: some are numerics (eg prices) and needs a numeric type column, others are in text format but with a numeric content (phone numbers with leading zeros).

When i try to export them, giving the right format to the value (char, int, float, etc..) it happens that cell values are converted without considering the initial format, the one i set.

This happens because tablib xlsx writer converts all values to unicode:

https://github.com/kennethreitz/tablib/blob/master/tablib/formats/_xlsx.py#L140

``` python
              try:
                if '\n' in col:
                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
                            '%s' % col, errors='ignore')
                        style = ws.get_style('%s%s' % (col_idx, row_number))
                        style.alignment.wrap_text
                    else:
                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
                            '%s' % col, errors='ignore')
                except TypeError:
                    ws.cell('%s%s'%(col_idx, row_number)).value = unicode(col)
```

so this cause openyxl to auto parse values with numeric content as a numeric cell:

https://github.com/kennethreitz/tablib/blob/master/tablib/packages/openpyxl3/cell.py#L219

``` python
       def data_type_for_value(self, value):
        """Given a value, infer the correct data type"""
        if value is None:
            data_type = self.TYPE_NULL
        elif value is True or value is False:
            data_type = self.TYPE_BOOL
        elif isinstance(value, (int, float)):
            data_type = self.TYPE_NUMERIC
        elif not value:
            data_type = self.TYPE_STRING
        elif isinstance(value, (datetime.datetime, datetime.date)):
            data_type = self.TYPE_NUMERIC
        elif isinstance(value, str) and value[0] == '=':
            data_type = self.TYPE_FORMULA
        elif self.RE_PATTERNS['numeric'].match(value):
            data_type = self.TYPE_NUMERIC
        elif value.strip() in self.ERROR_CODES:
            data_type = self.TYPE_ERROR
        else:
            data_type = self.TYPE_STRING
        return data_type
```

instead of converting all cell values to unicode,  wouldn't be possible to write cell values using openpyxl's 'set_value_explicit' checking the value format and passing the value using right openpyxl cell.TYPE_\* ?

vToMy:
+1 I get this problem for date types as well.

isedwards:
[...]

PR #309, opened 2017-09-11
Delegate type coercion to openpyxl

cristiano2lopes:
First of all, thank you for the great job on tablib, it helps a lot in our day to day job.

On my company we were having problems with the format of the columns exported on the format xlsx.

At the time of debug #252 seemed to be somewhat related with the problem we were observing (numeric fields appearing as text values and preventing spreadsheet app sums etc without a previous conversion inside the spreadsheet app ) so I decided to investigate a bit more. It seems that the default conversion of unicode of all the col values prevents openpyxl to detect the type of the value correctly, ending up treating all the values as text.

My initial approach, as presented in this pr for discussion, delegates the type detection and conversion to openpyxl, which seems like a sensible approach to me. (for what I've seen tablib already puts most of the export logic for xlsx on openpyxl's side). It fallsback to converting the value to unicode, and pass it again to openpyxl, if any data related conversion exception is raised (keeping compatibility, lowering the risk for existing codebases). 

I also noticed that no tests exist for this particular feature and I'am open to doing them (if there isn't any blocker already detected to doing them).

If this seems sensible I can proceed with some polishing (and maybe doing it for xls) and tests so it can be merged.

### DIFF
--- a/tablib/formats/_xlsx.py
+++ b/tablib/formats/_xlsx.py
@@ -120,33 +120,29 @@ def dset_sheet(dataset, ws, freeze_panes=True):
             if (row_number == 1) and dataset.headers:
                 # ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
                     # '%s' % col, errors='ignore')
-                ws.cell('%s%s'%(col_idx, row_number)).value = unicode(col)
                 style = ws.get_style('%s%s' % (col_idx, row_number))
                 style.font.bold = True
                 if freeze_panes:
                     # As already done in #53, but after Merge lost:
                     #  Export Freeze only after first Line
                     ws.freeze_panes = 'A2'
-                    
+
             # bold separators
             elif len(row) < dataset.width:
-                ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
-                    '%s' % col, errors='ignore')
                 style = ws.get_style('%s%s' % (col_idx, row_number))
                 style.font.bold = True
 
             # wrap the rest
             else:
                 try:
-                    if '\n' in col:
-                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
-                            '%s' % col, errors='ignore')
+                    str_col_value = unicode(col)
+                except TypeError:
+                    str_col_value = ''
+
+                    if '\n' in str_col_value:
                         style = ws.get_style('%s%s' % (col_idx, row_number))
                         style.alignment.wrap_text
-                    else:
-                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
-                            '%s' % col, errors='ignore')
-                except TypeError:
-                    ws.cell('%s%s'%(col_idx, row_number)).value = unicode(col)
+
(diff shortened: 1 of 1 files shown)


kennethreitz:
I'm afraid to touch any of this code, personally. Many people rely on it.

kennethreitz:
but yolo, right? 🤞

kennethreitz:
✨🍰✨

cristiano2lopes:
You're comments are making me nervous 😅. I will add some tests as soon as I can. Hopefully before a new release.

kennethreitz:
sounds good!

farzeni:
@cristiano2lopes i tried your repo and the patch seems to work

ragle:
@cristiano2lopes - this fixed the issue we were having with sorting / summing numerical columns as well in an xlsx export. Thanks for the fix, and thanks @kennethreitz for an awesome lib!

tablib/formats/_xlsx.py
```diff
@@ -120,33 +121,30 @@ def dset_sheet(dataset, ws, freeze_panes=True):
...
                 try:
-                    if '\n' in col:
-                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
-                            '%s' % col, errors='ignore')
-                        style = ws.get_style('%s%s' % (col_idx, row_number))
-                        style.alignment.wrap_text
-                    else:
-                        ws.cell('%s%s'%(col_idx, row_number)).value = unicode(
-                            '%s' % col, errors='ignore')
+                    str_col_value = unicode(col)
```
vissh: @cristiano2lopes `str_col_value` is not used.

tablib/formats/_xlsx.py
```diff
@@ -120,33 +121,30 @@ def dset_sheet(dataset, ws, freeze_panes=True):
...
                 except TypeError:
-                    ws.cell('%s%s'%(col_idx, row_number)).value = unicode(col)
+                    str_col_value = ''
 
+                    if '\n' in str_col_value:
```
vissh: To move to the left?

cristiano2lopes: The indentation is wrong i guess. The if should be outside the try except scope. And that would make the str_col_value used. This is probably ending with the wrapping being done only when we cannot obtain an unicode stirng from the col...

cristiano2lopes: Sorry about that.

isedwards:
For other people struggling with this, it's worth noting that the problem only affects the `.xlsx` method that produces Excel output.  Therefore a very quick workaround in the meantime is to bypass the method and use `openpyxl` to write the file directly.  Something simple like...

    import openpyxl

    def write_xlsx(tablib_dataset, filename):
        wb = Workbook()
        ws = wb.active
        ws.title = "Sheet 1"

        # Write the header row and make cells bold
        ws.append(tablib_dataset.headers)
        for cell in ws[1:1]:
            cell.font = Font(bold=True)

        for row in tablib_dataset:
            ws.append(row)

        wb.save(filename=filename)

PR #311, opened 2017-10-09
ujson presence no longer breaks tablib (resolves #297)

### LINKED ISSUE #297: JSON export broken when using `ujson`
When `ujson` is installed, [export_set](https://github.com/kennethreitz/tablib/blob/46102d4be718ee6fd6c003c1387df7c940e65b2c/tablib/formats/_json.py#L28) and [export_book](https://github.com/kennethreitz/tablib/blob/46102d4be718ee6fd6c003c1387df7c940e65b2c/tablib/formats/_json.py#L33) fail with

    # ...
    File "/usr/local/lib/python3.6/site-packages/tablib/formats/_json.py", line 30, in export_set
        return json.dumps(dataset.dict, default=date_handler)
    TypeError: 'default' is an invalid keyword argument for this function

This is because `ujson.dumps` [does not support the `default` kwarg and does not plan to in the future](https://github.com/esnme/ultrajson/issues/215).

rooterkyberian:
It seems to me like ujson support was broken from the day 1.
Fixing it would require recursively going through the structure and swapping datetime objects for `.isoformat()` equivalents, which would be surely slower than leaving it to cpython `json` library.

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -40,16 +40,6 @@ To download the full source history from Git, see :ref:`Source Control <scm>`.
 .. _zipball: http://github.com/kennethreitz/tablib/zipball/master
 
 
-.. _speed-extensions:
-Speed Extensions
-----------------
-
-You can gain some speed improvement by optionally installing the ujson_ library.
-Tablib will fallback to the standard `json` module if it doesn't find ``ujson``.
-
-.. _ujson: https://pypi.python.org/pypi/ujson
-
-
 .. _updates:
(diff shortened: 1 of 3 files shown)


rooterkyberian:
Failed only because `numpy` no longer supports python 3.3, which is yet to be excluded from travis builds (https://github.com/kennethreitz/tablib/pull/310)

iurisilvio:
Agreed. Merged! :cake:

Issue #324, opened 2018-02-01
Error exporting to XLSX: cell() takes at least 3 arguments (2 given)


dnmellen:
https://github.com/kennethreitz/tablib/blob/4c300e65a50eef72b91fa1909c9f68679723955e/tablib/formats/_xlsx.py#L122

I am having issues exporting to .xlsx: The upper line leads to the exception `cell() takes at least 3 arguments (2 given)`

In the openpyxl's code you can see the `cell` method expects the row and the column in different parameters:
https://bitbucket.org/openpyxl/openpyxl/src/d1a6b569dc46d66d84978826f87c9a1ff1141b25/openpyxl/worksheet/worksheet.py?at=default&fileviewer=file-view-default#worksheet.py-273

mjpieters:
This is a change in openpyxl 2.5.0, downgrading to 2.4.9 should fix this for now.

The arguments were made mandatory in [this revision](https://bitbucket.org/openpyxl/openpyxl/commits/d66c74fdd55ac699865ee5f7dff9bd483c2590fb).

appletonwebb:
I'd be great if this was fixed in the next release.

gbataille:
I think this is fixed by https://github.com/kennethreitz/tablib/pull/296 that has been submitted a while ago and that is waiting for action/comment

It's unclear what should be done on it.

in the meantime, maybe you should peg the tablib dependency to openpyxl to 2.4.9

n6151h:
Downgrading to 2.4.9 isn't an option.  Code that uses later versions breaks when using 2.4.9.

lwm:
Affected by this unfortunately.

I'll take a stab at https://github.com/kennethreitz/tablib/issues/324#issuecomment-362809550, thanks @mjpieters. 

Any fix coming for this?

brad:
Works on 2.4.11 as well

iurisilvio:
Fixed.

PR #325, opened 2018-02-01
Changes the way to get a cell from a xlsx file

dnmellen:
Relates to issues #324 and #324

Solves the exception `cell() takes at least 3 arguments (2 given)` when exporting to .xlsx


manelclos:
@dnmellen there is no version check for openpyxl, is this change backwards compatible with say 2.4.x versions?

iurisilvio:
Closing in favor of #296. Same change, already merged. Thanks! :cake:

PR #332, opened 2018-06-18
Fix: Circular reference detected error

bratomes:
### Problem
I got the error `ValueError: Circular reference detected` when trying to get a [Record instance representation](https://github.com/kennethreitz/records/blob/v0.5.2/records.py#L45) of records lib.

### Solution
My table has a uuid type field. This type is not JSON serializable, so needs to be handled (converted to str).

Maybe this fix https://github.com/kennethreitz/records/issues/95.

### Error scenario
Create database with a uuid field
```sql
create table bar (                                                      
    id serial primary key,
    foo_id uuid
);
```
Insert data
```sql
insert into bar (foo_id) values ('30fcb0c7-a836-446c-9c88-411800c0c818')
```
Use records lib and get the error
```python
In [1]: import records
In [2]: db = records.Database('postgres://...')
In [3]: rows = db.query('select * from bar')
In [4]: rows[0]
...
ValueError: Circular reference detected
```

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,4 @@ Patches and Suggestions
 - Mathias Loesch
 - Tushar Makkar
 - Andrii Soldatenko
+- Bruno Soares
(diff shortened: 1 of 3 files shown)


dschep:
Hey @kennethreitz, any chance of this change being released to PyPI?

PR #336, opened 2018-07-15
reStructuredText

kaapstorm:
Adds support for exporting datasets and databooks as reStructuredText.

This is useful for using data from a range of sources in documentation.

```python
>>> data = Dataset().load(open('data.csv').read())
>>> data.export(rst)
=====  =====  =====
  A      B    A and
                B  
=====  =====  =====
False  False  False
True   False  False
False  True   False
True   True   True 
=====  =====  =====
```

### DIFF
--- a/tablib/compat.py
+++ b/tablib/compat.py
@@ -22,6 +22,8 @@ except ImportError:
 
 if is_py3:
     from io import BytesIO
+    from itertools import zip_longest as izip_longest
+    from statistics import median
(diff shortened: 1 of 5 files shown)


kaapstorm:
(Travis failure seems unrelated. Build attempts to install numpy 1.13.1 (not a requirement of this PR) on Python 3.3, but it requires Python >= 3.4.)

Issue #207, opened 2015-11-07
Streaming responses


aidanlister:
Can we use tablib to stream responses, eg have dataset.csv return a generator?

kennethreitz:
This is not currently planned.

jarshwah:
Sorry to rehash, but would you accept a patch that implemented this support?

iurisilvio:
I'm +1 for it, but it needs some design decisions. Which operations we'll be able to support for streaming imports and exports?

I know openpyxl has optimized reader/writer (http://openpyxl.readthedocs.io/en/default/optimized.html) and it is really useful.

Reopening the issue for discussion. I think it is possible to make it in tablib.

ZuluPro:
Hello @iurisilvio @jarshwah @kennethreitz 

A lot of formats use file-like object but `export_set` returns `stream.getvalue()` to get a string.
So my proposition is the following:
- Add the function `export_stream_set` to the possible ones
- Update `export_set` to use the `export_stream_set` and launch `.getvalue()` to get a string.

I think we could gain in features just by split code in several functions.
And I really need to be able to work with streams :D

What do you think about ?

ZuluPro:
POC here: #337

PR #337, opened 2018-08-08
Added stream to CSV

ZuluPro:
Hello
Here's a proof of concept of what I talk in #207 .
This PR just add the possibility to get a stream or its content.
I didn't change the API, just added `seek(0)`, to have a file ready to use.

I use the same code here https://github.com/django-import-export/django-import-export/pull/821/files#diff-9c335e3895cab39c4af9e510b328c2f0R157
It's a Django app for import/export from DB.

To have a good integration we need to get file-like object.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,4 @@ junit-py27.xml
 
 # pyenv noise
 .python-version
+tablib.egg-info/*
--- a/tablib/formats/_csv.py
+++ b/tablib/formats/_csv.py
@@ -13,8 +13,8 @@ extensions = ('csv',)
 DEFAULT_DELIMITER = unicode(',')
 
 
-def export_set(dataset, **kwargs):
-    """Returns CSV representation of Dataset."""
+def export_stream_set(dataset, **kwargs):
+    """Returns CSV representation of Dataset as file-like."""
     stream = StringIO()
 
     kwargs.setdefault('delimiter', DEFAULT_DELIMITER)
@@ -24,6 +24,13 @@ def export_set(dataset, **kwargs):
     for row in dataset._package(dicts=False):
         _csv.writerow(row)
 
+    stream.seek(0)
+    return stream
+
+
+def export_set(dataset, **kwargs):
+    """Returns CSV representation of Dataset."""
+    stream = export_stream_set(dataset, **kwargs)
     return stream.getvalue()
 
 
--- a/test_tablib.py
+++ b/test_tablib.py
@@ -13,7 +13,7 @@ from uuid import uuid4
 import tablib
 from tablib.compat import markup, unicode, is_py3
 from tablib.core import Row
-from tablib.formats import csv as csv_format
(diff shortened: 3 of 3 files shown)


timofurrer:
Could you please rebase onto master? Thanks :tada:

ZuluPro:
Hey @timofurrer

I rebased from master.

claudep:
Could that feature be tested to avoid regressions?

ZuluPro:
@claudep 
I'm gonna write tests of course ;)
But can you me that this will be ± the final API for this feature ?

claudep:
I'm no maintainer, but if a maintainer asked you to rebase, that's a good sign the API is fine!

timofurrer:
> the final API for this feature ?

There won't be a *final API* - but we'll adjust the versioning accordingly to API breaks, so we don't break your code.
However, for now it looks good to me 🎉

ZuluPro:
@claudep @timofurrer 
I added a test

misli:
In order to do a real stream, there also need to be a streaming input and streaming processing. Recently I've implemented following StreamingDataset, which is (in cooperation with custom streaming formats) capable of working with iterator (generator) as data. It allows me to export large database tables with constant memory requirements.
```python
class _SubscriptableIterable:
    def __init__(self, iterable):
        self._iter = iter(iterable)
        self._first_item = next(self._iter)

    def __getitem__(self, index):
        if index == 0:
            return self._first_item
        raise IndexError

    def __iter__(self):
        yield self._first_item
        yield from self._iter


class StreamingDataset(Dataset):
    def __init__(self, data=None, **kwargs):
        super().__init__(**kwargs)

        if data is not None:
            self.set_data(data)

    def set_data(self, data):
        self._data = _SubscriptableIterable(map(Row, data))

    def __iter__(self):
        return self

    def __next__(self):
        return next(self._data)

    def __repr__(self):
        try:
            return "<%s streaming dataset>" % (self.title.lower())
        except AttributeError:
            return "<streaming dataset>"

    def _apply_formatters(self):
        if not self._formatters:
            yield from self._data
        for row in self._data:
            for col, callback in self._formatters:
                try:
                    if col is None:
                        for j, c in enumerate(row):
                            row[j] = callback(c)
                    else:
                        row[col] = callback(row[col])
                except IndexError:
                    raise InvalidDatasetIndex
                yield row

    def _package(self, dicts=True, ordered=True):
        """Packages Dataset into lists of dictionaries for transmission."""

        if ordered:
            dict_pack = OrderedDict
        else:
            dict_pack = dict

        data = self._apply_formatters()

        if self.headers:
            if dicts:
                data = (dict_pack(zip(self.headers, data_row)) for data_row in data)
            else:
                data = chain([self.headers], data)

        return data
```

harkabeeparolus:
If you want to stream large sets of data, I would recommend looking into petl, which is also a pure Python tabular data library.

https://github.com/petl-developers/petl

Issue #341, opened 2018-11-25
Loading big CSV files is slow due to sniffing


rivol:
I'm using tablib to load a large-ish (2.5MB, 3k rows) CSV file.

tablib's format detector uses `csv.Sniffer` to see if the file is CSV, but passes the entire data stream to the sniffer. In my case this causes the format detection to take about 5 minutes. The impression I'm getting from [the official CSV module docs](https://docs.python.org/3/library/csv.html#csv.Sniffer) is that `Sniffer` should be passed only a small sample of data, e.g. 1kb, not the whole file.

I propose passing only the first 1kb of data to the `Sniffer` at https://github.com/kennethreitz/tablib/blob/5a359ba4def1611b34df003bec58439a125d4e56/tablib/formats/_csv.py#L54 
This is same amount as shown in the official docs.

Can do PR if this approach sounds good.

claudep:
Still motivated to propose a PR?

claudep:
Fixed by 8f09789d4 (the commit message wrongly referenced #342).

PR #342, opened 2018-12-12
Delegate type coercion to openpyxl

heykarimoff:


heykarimoff:
Found useless))

claudep:
The above commit should have referenced #341, not #342.

PR #270, opened 2017-01-12
Import ascii characters not valid with unicode literals

browniebroke:
I had the same bug as #215 so I started working on a fix. It looks like it's a problem is around unicode literals.

- [x] Test case to reproduce the bug
- [x] Provide a patch

PR #350, opened 2019-03-02
Import ascii characters not valid with unicode literals - updated

### LINKED ISSUE #215: Import ascii characters not valid.
I tried to add data to the tablib Dataset. I use django import export to do so. Everything goes well with the data until it is in the dataset. Once it is added and I try to read the data I get the following error. 

```
UnicodeEncodeError encountred while trying to read file: 'ascii' codec can't encode character u'\xc6' in position 241: ordinal not in range(128)
```

It is about the text Ævar. This name will fail when you try to read the data from the Dataset.

browniebroke:
Replaces #270 and fixes #215


browniebroke:
@timofurrer here is the updated version.

timofurrer:
Looks good! Thanks for updating and sorry, for the long delay! 🎉

PR #358, opened 2019-03-09
Removed distutils fallback

claudep:
As of https://github.com/kennethreitz/setup.py/blob/master/setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -5,10 +5,8 @@ import os
 import re
 import sys
 
-try:
-    from setuptools import setup
-except ImportError:
-    from distutils.core import setup
+from setuptools import setup
+
 
 if sys.argv[-1] == 'publish':
     os.system("python setup.py sdist upload")


claudep:
Then you won't be able to install this package anyway as several of its dependencies are also directly importing setuptools.setup without fallback. I think the time caring for a missing setuptools package is over now.

chfw:
Here's the reference to the issue: https://github.com/moremoban/pypi-mobans/issues/66.

The main concern is for embedded device and docker builds where size matters.

claudep:
I see, however I'm afraid unless a PEP changes that, the current good practice is still to directly import setuptools.setup.
See https://docs.python.org/3/library/distutils.html which is some official endorsement of setuptools.

chfw:
Thanks for further elaboration and we now understand both sides of the opinion. I have no objections for the merge.

timofurrer:
Totally agree! Thanks 🎉

PR #360, opened 2019-03-09
Refs #273 - Replaced vendored markup lib by dependency

claudep:


jdufresne:
The file `NOTICE` could be removed with this change

https://github.com/kennethreitz/tablib/blob/733d77ad1ed393869825f74803781d9b9873782e/NOTICE#L1-L6

claudep:
Good catch!

ParthS007:
@claudep Can you please resolve conflicts. Thanks for the PR!

claudep:
Sure, done.

jayvdb:
I noticed the new repo for MarkupPy doesnt contain any code - https://github.com/tylerbakke/MarkupPy/issues/3 , but I guess it wasnt being modified much anyway as it is a 1:1 replacement for the vendored code in tablib.

IMO `tablib/compat.py` and `tablib/packages/__init__.py` should still have imports in them, importing the external libraries, to avoid breaking the interface of `compat` which users may be relying on.

Also it would be good if the setup.py reference has the correct capitalisation, MarkupPy, but not important as most setup.py inspection tools now get the requirements from requirements.txt .

PR #361, opened 2019-03-09
Unpin transient dependencies in requirements.txt

jdufresne:
The project is expected to work with the all versions of dependencies as
specified by dependency ranges, not just a single pinned version. Stop
overspecifying them.


jdufresne:
It comes from `setup.py`. It is the lowest version allowed by the project:

https://github.com/kennethreitz/tablib/blob/733d77ad1ed393869825f74803781d9b9873782e/setup.py#L37

jdufresne:
Rebased to resolve conflicts

PR #362, opened 2019-03-12
Use statistics backport

jayvdb:
There is a de-vendoring process happening, https://github.com/kennethreitz/tablib/issues/273 , and then a new one appears in the latest release.


claudep:
As for me, I wouldn't mind keeping this code for the remaining duration of Python 2 support, which shouldn't be too long anyway. A dependency for 8 lines of code is a bit overkill, but let the maintainers decide.

jayvdb:
Well its a dependency only for people on Python 2, and it means this library is able to use anything from Python 3 statistics , and nobody can modify it - if there are bugs in it, they should be fixed elsewhere.  That avoids the problems encountered with dbfpy, where there are local modifications.

Also the license is different, so vendoring it means this library needs to carry an additional license, which adds complexity to users who need to do license checks before using a library.

timofurrer:
Well, the question here is, if it's worth to use the backport just for `statistics.median` function?
The currently "vendored" *statistics.py* only contains the `median()` function - which is not even the exact same as in cpython master.

claudep:
Let this small vendored code die with Python 2.

PR #365, opened 2019-05-16
Avoid DeprecationWarning due to invalid escapes in docstrings

bluetech:
Will become a SyntaxError in Python 3.8:
https://bugs.python.org/issue32912

### DIFF
--- a/tablib/core.py
+++ b/tablib/core.py
@@ -142,7 +142,7 @@ class Dataset(object):
 
         data = tablib.Dataset(*data, headers=headers)
 
-    :param \*args: (optional) list of rows to populate Dataset
+    :param \\*args: (optional) list of rows to populate Dataset
     :param headers: (optional) list strings for Dataset header row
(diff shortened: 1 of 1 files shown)


hugovk:
On https://bugs.python.org/issue32912, they decided to wait a little longer, until Python 3.9, to raise a `SyntaxError` (https://github.com/python/cpython/pull/15142), so this will still be a  `DeprecationWarning` in 3.8.

Still, now is as good time to fix the `DeprecationWarning`!

hugovk:
Python 3.8.0 final is due out on **October 14th**, with the release candidate out now.

* https://discuss.python.org/t/currently-working-on-the-release-of-python-3-8-0rc1/2411

Also, Python 3.9 development has begun, and the first alpha is also due out on **October 14th**.

* https://www.python.org/dev/peps/pep-0596/

See also https://github.com/jazzband/tablib/pull/382.

PR #374, opened 2019-08-11
Force default_flow_style for pyyaml safe_dump

aerostitch:
Pyyaml 5.1 changed their default_flow_style for their *dump methods.
This patch is here to maintain a consistent behavior in the output before and after upgrading pyyaml to 5.1.
You can see more details in https://bugs.debian.org/934271

### DIFF
--- a/src/tablib/formats/_yaml.py
+++ b/src/tablib/formats/_yaml.py
@@ -13,12 +13,12 @@ class YAMLFormat:
     def export_set(cls, dataset):
         """Returns YAML representation of Dataset."""
 
-        return yaml.safe_dump(dataset._package(ordered=False))
+        return yaml.safe_dump(dataset._package(ordered=False), default_flow_style=None)
 
     @classmethod
     def export_book(cls, databook):
         """Returns YAML representation of Databook."""
-        return yaml.safe_dump(databook._package(ordered=False))
+        return yaml.safe_dump(databook._package(ordered=False), default_flow_style=None)
 
     @classmethod
     def import_set(cls, dset, in_stream):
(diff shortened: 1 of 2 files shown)


aerostitch:
Updated the PR with the latest changes.

claudep:
Do you think some test could be added to show the difference?

aerostitch:
Hey @claudep ,

I'll have a look but not sure if that can really be caught in unit tests. But basically using this simple script:
```python
import tablib

headers = ('first_name', 'last_name')
data = [('John', 'Adams'), ('George', 'Washington')]
data = tablib.Dataset(*data, headers=headers)

export_yaml = data.export('yaml')
print(export_yaml)
```

Before pyyaml 5.1 (and after that if you add this patch), the output will be:
```
$ python3 bla.py
- {first_name: John, last_name: Adams}
- {first_name: George, last_name: Washington}
```

But after pyyaml 5.1, if you don't add this patch you will end up with the following output which breaks the continuity between versions:
```
$ python3 bla.py
- first_name: John
  last_name: Adams
- first_name: George
  last_name: Washington
```

Joseph

aerostitch:
Hey @claudep ,
I've added the missing unit test to check the output of the export format in yaml.
It was there for other formats but not this one.
Let me know your thoughts on this.
Thanks,
Joseph

claudep:
Awesome, thanks! I will merge that soon.
What would be still greater is to allow the `export_*` methods to accept the kwargs for `safe_dump` (aka `dump_all`), like we are doing for the `csv` module. But that would be the subject of a separate issue.

PR #382, opened 2019-10-03
Test on Python 3.8 beta

hugovk:
Python 3.8.0 final is due out on **October 14th**, with the release candidate out now.

* https://discuss.python.org/t/currently-working-on-the-release-of-python-3-8-0rc1/2411

It's a good idea to test it out now, to avoid any big surprises in two weeks.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,7 @@ matrix:
     - python: 3.5
     - python: 3.6
     - python: 3.7
+    - python: 3.8-dev
 install:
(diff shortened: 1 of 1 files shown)


hugovk:
The first 3.8-dev build is slow because it's building dependencies:

* https://travis-ci.org/jazzband/tablib/builds/592953231

But subsequent builds are fast because they're cached:

* https://travis-ci.org/hugovk/tablib/builds/592951935

Issue #370, opened 2019-07-08
IllegalCharacterError raised when exporting xlsx


MohiniLimbodia:
Using tablib version 0.13.0

Steps to reproduce:
```py
from tablib import Dataset
data=Dataset((u'\x1f',),)
data.export('xlsx')
```

```
IllegalCharacterErrorTraceback (most recent call last)
<ipython-input-1-9b9694f739ff> in <module>()
      1 from tablib import Dataset
      2 data=Dataset((u'\x1f',),)
----> 3 data.export('xlsx')

/srv/jupyter/local/lib/python2.7/site-packages/tablib/core.pyc in export(self, format, **kwargs)
    466             raise UnsupportedFormat('Format {0} cannot be exported.'.format(format))
    467 
--> 468         return export_set(self, **kwargs)
    469 
    470     # -------

/srv/jupyter/local/lib/python2.7/site-packages/tablib/formats/_xlsx.pyc in export_set(dataset, freeze_panes)
     41     ws.title = dataset.title if dataset.title else 'Tablib Dataset'
     42 
---> 43     dset_sheet(dataset, ws, freeze_panes=freeze_panes)
     44 
     45     stream = BytesIO()

/srv/jupyter/local/lib/python2.7/site-packages/tablib/formats/_xlsx.pyc in dset_sheet(dataset, ws, freeze_panes)
    145                         cell.value = unicode('%s' % col, errors='ignore')
    146                 except TypeError:
--> 147                     cell.value = unicode(col)

/srv/jupyter/local/lib/python2.7/site-packages/openpyxl/cell/cell.pyc in value(self, value)
    292     def value(self, value):
    293         """Set the value and infer type and display options."""
--> 294         self._bind_value(value)
    295 
    296     @property

/srv/jupyter/local/lib/python2.7/site-packages/openpyxl/cell/cell.pyc in _bind_value(self, value)
    195 
    196         elif isinstance(value, STRING_TYPES):
--> 197             value = self.check_string(value)
    198             self.data_type = self.TYPE_STRING
    199             if len(value) > 1 and value.startswith("="):

/srv/jupyter/local/lib/python2.7/site-packages/openpyxl/cell/cell.pyc in check_string(self, value)
    158         value = value[:32767]
    159         if next(ILLEGAL_CHARACTERS_RE.finditer(value), None):
--> 160             raise IllegalCharacterError
    161         return value
    162 

IllegalCharacterError:
```

leonardoarroyo:
[...]

PR #386, opened 2019-10-03
Fixes #314 - Delegate type coercion to openpyxl

### LINKED ISSUE #314: xlsx casting all types to string | unicode
```python
head = ('integer', 'float', 'text', 'integer as text')
data = [(123, 3.14, 'hello world', '987',),]
data = tablib.Dataset(*data, headers=head)
f = open('types.xlsx', 'wb')
f.write(data.xlsx) # works correctly with data.xls
```

**Expected behaviour:**

Exported file would have values in the same type as defined above.
Eg. integer => integer,  float => float, string => string

**What is happening**

All types are casted to strings
integer => string,  float => string, string => string

This is a problem when people what to do some counting operation with the resulting `xlsx` as it will break, and needs to be changed inside the document.

**Where and when is the origin**

* Introduced in `version  0.11.5` in [commit](https://github.com/kennethreitz/tablib/pull/221/files#diff-e084e40878a6d2614b831e70c87fb140R142) by adding unicode() around `col`
* imho a silent BC, as it changes functionality

Internally unicode function return a string object, or for python3 compatibility is casted to string object:
```python
# inside compat.py
if is_py3:
    unicode = str;
```

**Fix**

Leave handling of types for `openpyxl` and remove unicode wraps around values.

This fix was proposed and merged already in #309, but it was done again develop branch that is quite outdated and not compatible with python3.
Also when exporting to `xls` it works correctly as unicode wrapping wasn't added here.

claudep:
Thanks Cristiano Lopes for the initial patch.

### DIFF
--- a/test_tablib.py
+++ b/test_tablib.py
@@ -782,6 +782,25 @@ class TSVTests(BaseTestCase):
         self.assertEqual(tsv, self.founders.tsv)
 
 
+class XLSXTests(BaseTestCase):
+    def test_xlsx_import_set(self):
+        data.append(('string', 42, 21.55))
+        data.headers = ('string', 'integer', 'float')
+        _xlsx = data.xlsx
+        data.xlsx = _xlsx
+        self.assertEqual(data.dict[0]['string'], 'string')
+        self.assertEqual(data.dict[0]['integer'], 42)
+        self.assertEqual(data.dict[0]['float'], 21.55)
+
+    def test_xlsx_wrong_char(self):
+        """Bad characters are not silently ignored. We let the exception bubble up."""
+        from openpyxl.utils.exceptions import IllegalCharacterError
+
+        with self.assertRaises(IllegalCharacterError):
+            data.append(('string', b'\x0cf'))
+            data.xlsx
+
+
 class JSONTests(BaseTestCase):
     def test_json_format_detect(self):
         """Test JSON format detection."""
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -4,6 +4,8 @@
 
 - The project is now maintained by the Jazzband team, https://jazzband.co
 - Fixed `DataBook().load` parameter ordering (first stream, then format).
+- Fixed a regression for xlsx exports where non-string values were forced to
+  strings (#314).
 - Added search to all documentation pages
 - Open xlsx workbooks in read-only mode (#316)
 - Unpin requirements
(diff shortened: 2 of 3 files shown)


test_tablib.py
```diff
@@ -782,6 +782,22 @@ def test_tsv_export(self):
...
+        data.headers = ('string', 'integer', 'float')
+        _xlsx = data.xlsx
+        data.xlsx = _xlsx
+        self.assertEqual(data.dict[0]['string'], 'string')
+        self.assertEqual(data.dict[0]['integer'], 42)
+        self.assertEqual(data.dict[0]['float'], 21.55)
+
+    def test_xlsx_wrong_char(self):
+        """A bad character should not crash the export"""
+        data.append(('string', b'\x0cf'))
```
claudep: @hugovk That test currently fails and I'd like to know your opinion. `openpyxl` is refusing to store some chars in a certain range:
https://bitbucket.org/openpyxl/openpyxl/src/15df72eae2f93bedc0c0c0cc8753b9107cc82d41/openpyxl/cell/cell.py#lines-70

Should we let the error break the export, or catch it and replace the offended chars by some value (empty string, '#N/A'?) I tend to favor letting the exception crash the export (and change the test accordingly).

hugovk: Right now it's raising `IllegalCharacterError`:

```
  File "test_tablib.py", line 798, in test_xlsx_wrong_char
    data.xlsx
  File "/home/travis/build/jazzband/tablib/tablib/formats/_xlsx.py", line 39, in export_set
    dset_sheet(dataset, ws, freeze_panes=freeze_panes)
  File "/home/travis/build/jazzband/tablib/tablib/formats/_xlsx.py", line 142, in dset_sheet
    cell.value = unicode(col, errors='ignore')
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/openpyxl/cell/cell.py", line 239, in value
    self._bind_value(value)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/openpyxl/cell/cell.py", line 215, in _bind_value
    value = self.check_string(value)
  File "/home/travis/virtualenv/python3.7.1/lib/python3.7/site-packages/openpyxl/cell/cell.py", line 182, in check_string
    raise IllegalCharacterError
openpyxl.utils.exceptions.IllegalCharacterError
```

I think that's probably the right thing to do, it will let people know what the problem is and let them react appropriately (change the character, delete it, etc.).

@cristiano2lopes, what do you think?

claudep: #370 is related.

claudep:
I think I will push the patch in the current state, and we can continue discussing the illegal characters exception in #370.

PR #391, opened 2019-10-04
Fix missing comma in setup.py

peymanslh:
Missing comma in `setup.py` file causes an error while running the tests:
```
ERROR: Could not find a version that satisfies the requirement markuppyxlrd (from tablib==0.13.0) (from versions: none)
ERROR: No matching distribution found for markuppyxlrd (from tablib==0.13.0)
```
pip can't install `markuppyxlrd` (markuppy, xlrd)


claudep:
Wow, good catch, thanks!

Issue #256, opened 2016-10-16
Framework for adding input/output formats


gwerbin:
Right now, the list of formats is more or less hardcoded into the `formats.available` tuple, and it's not straightforward to extend (at least not to me).

It would be great if you could subclass an abstract `Formatter` class, and then either have it magically register (using some kind of metaclass magic, I suppose) with the available formats list or introduce a function and/or class decorator to explicitly register the new format.

Is that something you'd be interested in working on in the near term? I'd be happy to contribute some of my own time for it.

iurisilvio:
I'm looking for a good solution to this issue and really want it done.

Ideas?

gwerbin:
Ok. Unfortunately I've gotten a lot busier recently and won't have time to look at this until well into the new year. I'll post an update when I've had a chance to size up the task.

iurisilvio:
No problem. I'm still trying to decide the best API for it.

Based on your suggestions:

 * I don't like the subclass magic register
 * I like the class decorator idea

I tried to make formats available per dataset/databook instance, but it creates some complicated problems.

Also, I started a solution this way:

```python
from tablib import formats
formats.register("zip", MyFormatter)
```

We can do the same thing with decorators.

I'll finish my work and push it to a branch to receive some feedback.

gwerbin:
Looks like we both might have forgotten about this a bit. I don't have any time now to work on this unfortunately.

claudep:
The PR #395 converts all formats to classes, with a registration pattern. This should allow to easily add custom formats. I think I was able to not (or minimally) break the current tablib API. I would be very interested to get input about the chosen approach.

iurisilvio:
I don't think you need to enforce classes there. If the `.register` receive the object instead of the class (or at least check if it is a class or a module), the current way would still work.

claudep:
You are probably right, but I think using classes gives much more power to the customization ability, and can save code too. See for example the new  `tablib/formats/_tsv.py`. You can easily subclass any tablib format, and just customize the attribute or method you need.
I will still explore your suggestion to accept an instance instead of a class.

hugovk:
@gwerbin @iurisilvio @LaundroMat Please could you review PR #395? Thank you!

claudep:
I'd like to go forward with this issue. I would have much appreciated more reviews, but I guess we can still polish this refactoring in master before the next release. Any opposition?

claudep:
This is now in master.

Issue #390, opened 2019-10-04
Drop support for Python 2


hugovk:
Python 2 reaches end-of-life on **2020-01-01**.

See https://python3statement.org for a list of projects which have pledged to drop support for Python 2 on or before its EOL.

I'd like to suggest making a release soonish with the recent changes that still supports Python 2.7, and then drop it.

That'll allow newer features of Python to be used, and also some vendored code  can be removed (eg. https://github.com/jazzband/tablib/pull/362#issuecomment-538104111) and deal with some other issues (eg. https://github.com/jazzband/tablib/pull/268#issuecomment-473651820).

claudep:
+1. What would you suggest for the next version numbering, 0.13.1?

hugovk:
I don't know the version scheme being used so far, but if we look at the popular [SemVer](https://semver.org):

> 4. Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.
>
> 5. Version 1.0.0 defines the public API. The way in which the version number is incremented after this release is dependent on this public API and how it changes.

...

> **How do I know when to release 1.0.0?**
>
> If your software is being used in production, it should probably already be 1.0.0. If you have a stable API on which users have come to depend, you should be 1.0.0. If you’re worrying a lot about backwards compatibility, you should probably already be 1.0.0.

Tablib was downloaded nearly 300,000 times last month, I think it's about time for 1.0.0 :)

claudep:
I'm not very comfortable releasing 1.0.0 right now, I'd rather stick with 0.13.1 or 0.14.0 and release 1.0.0 after dropping Python 2. Thoughts?

hugovk:
Agreed. (I mixed up the *next* release with the drop release.)

SemVer says:

* For 0.y.z releases, anything goes.

* For 1+.y.z releases, bump the patch number x.y.Z when there's bugfixes.

* For 1+.y.z releases, bump the minor number x.Y.z when there's new features (and optionally also bugfixes).

I'd probably go for 0.14.0.

peymanslh:
Hello @hugovk @claudep 
Can I work on this issue and send a pull request?

claudep:
Fine for me. It will probably need some rebasing at the time we're ready to commit it.

peymanslh:
I sent a pr: #394 
@claudep

claudep:
PR #394 was merged "manually".

PR #395, opened 2019-10-05
Refs #256 - Implement class-based formats

claudep:
This allows to extend Tablib with new formats far more easily.

### DIFF
--- a/docs/development.rst
+++ b/docs/development.rst
@@ -90,32 +90,36 @@ Tablib features a micro-framework for adding format support.
 The easiest way to understand it is to use it.
 So, let's define our own format, named *xxx*.
 
-1. Write a new format interface.
+From version 1.0, Tablib formats are class-based and can be dynamically
+registered.
 
-    :class:`tablib.core` follows a simple pattern for automatically utilizing your format throughout Tablib.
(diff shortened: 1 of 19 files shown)


claudep:
Failure on Python 2.7 is expected. This will not be pushed before the Python 2.7 support is dropped.

docs/development.rst
```diff
@@ -90,32 +90,36 @@ Tablib features a micro-framework for adding format support.
 The easiest way to understand it is to use it.
 So, let's define our own format, named *xxx*.
 
-1. Write a new format interface.
+From version 0.15.0, Tablib formats are class-based and can be dynamically
```
hugovk: Will the next one be 1.0, because Python 2 has been dropped?

Re: https://github.com/jazzband/tablib/issues/390#issuecomment-538331435

claudep: I hesitated, should we pass to 1.0 after such an important change? I also plan to make import functions accept Path or filelike objects in input (local patch ready), so I'd like to have one more 0.x release for those changes. But it's not a strong will, so I'm ready to use 1.0 if you think it's time.

claudep: OK, changed.

PR #394, opened 2019-10-04
Drop python 2 support

peymanslh:
**Issue:** [#390 ](https://github.com/jazzband/tablib/issues/390)

- Remove support python 2 from the docs, `requirements.txt`, `travis-ci`, `setup.py`, `tox.ini`
- Replace `unicode` with `str`
- Remove `dbfpy` folder and rename `dbfpy3` to `dbfpy`
- Remove `compat` file and remove python2 packages from dependency

PR #396, opened 2019-10-05
Deprecate Python 2.7 support

aleksihakli:
Edit: seems like this is a duplicate of #394 :)


claudep:
Yes, it is. Sorry for your double work.

PR #397, opened 2019-10-05
Test with pytest and send coverage to Codecov

hugovk:
Use pytest to test with coverage, and send the coverage to Codecov.

This will help ensure PRs maintain coverage and that new code is tested.

Coverage is currently 79%, which is pretty good.

Here's an example report on Codecov:

* https://codecov.io/gh/jazzband/tablib/tree/86a1b1dbe3e13b35a2133123ddfd212a44772daa/tablib

### DIFF
--- a/.coveragerc
+++ b/.coveragerc
@@ -1,0 +1,10 @@
+# .coveragerc to control coverage.py
+
+[report]
+# Regexes for lines to exclude from consideration
+exclude_lines =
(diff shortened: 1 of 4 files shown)


claudep:
Thanks. If we can continue using plain Python unit testing in parallel, I'm fine.

hugovk:
Yep, you can still run `python test_tablib.py` as before.

PR #398, opened 2019-10-18
Add project release config and use Travis build stages.

jezdez:
Refs #378.


.travis.yml
```diff
@@ -1,20 +1,32 @@
...
+    - pip install -r requirements.txt
+    - pip install -U pytest pytest-cov
+    after_success:
+    - pip install -U codecov
+    - codecov
+  - stage: deploy
+    script: skip
+    deploy:
+      provider: pypi
+      user: jazzband
```
hugovk: PyPI now supports uploading via API token:

https://pythoninsider.blogspot.com/2019/07/pypi-now-supports-uploading-via-api.html

> **Why:** These API tokens can **only** be used to upload packages to PyPI, and not to log in more generally. This makes it safer to automate package upload and store the credential in the cloud, since a thief who copies the token won't also gain the ability to delete the project, delete old releases, or add or remove collaborators. And, since the token is a long character string (with 32 bytes of entropy and a service identifier) that PyPI has securely generated on the server side, we vastly reduce the potential for credential reuse on other sites and for a bad actor to guess the token.

Something to consider?

jezdez: Not really since this does the release to the Jazzband package index (a staging areas of sorts) so that project leads can confirm each release. You'll get an emails once it's done. See https://jazzband.co/about/releases for more info.

.travis.yml
```diff
@@ -1,20 +1,35 @@
+sudo: false
```
hugovk: `sudo` can be removed, it no longer has any effect:

* https://blog.travis-ci.com/2019-10-07-multi-cpu-architecture-support

jezdez: Thanks!

jezdez:
@hugovk @claudep To fully automate the release process it's strongly recommended to use [setuptools-scm](https://github.com/pypa/setuptools_scm/) to get the version string from Git during the build process (and effectively allowing to do releases simply by creating a Git tag). That would remove the need to update Python files to bump up the version. This strategy has worked very well with many other Python projects including Jazzband projects. Just want to give you a heads-up that this will contain this change as well.

hugovk:
+1 for setuptools-scm, it makes versioning and releases _much_ easier.

One important thing to note: by default it includes everything from Git in the package, making the releases bigger. If we don't want all that, we'll need to explicitly exclude/prune other files/dirs.

For example: https://github.com/hugovk/pypistats/pull/63

jezdez:
@hugovk I'm glad you approve! That said, from the Python packaging conceptual side speaking, the purpose of source distributions is to ship all files that developers may want to use to inspect a file, including test files etc. So while the dist files may be bigger, they would also be matching the expectation of users.

src/tablib.egg-info/PKG-INFO
```diff
@@ -0,0 +1,464 @@
+Metadata-Version: 2.1
```
hugovk: Should the `*.egg-info` dir be Git ignored and removed from source control?

For example:

* https://github.com/github/gitignore/blob/21419e391a78f2487340b3b1240b988aaf15b54f/Python.gitignore#L25

hugovk:
Sure, that's fine and makes sense! It's probably only a few KB bigger anyway, just wanted to flag it up as it has surprised some people in the past :)

jezdez:
Yep, much appreciated that you're raising this.

claudep:
Wow, many thanks for pushing that! Packaging stuff has always been a weak point for me, so I'm glad you're taking the lead on that part.

Issue #293, opened 2017-08-03
Unpickled Dataset results in unusable CSV and JSON exports


woozyking:
```TypeError: <csv/json and probably more>() takes 0 positional arguments but 1 was given```

This happens after unpickling a pickled `Dataset` instance. More specifically, the pickling/unpickling process is through `rq` (http://python-rq.org/)

`.dict` would still work, and direct `repr/str` of the Dataset instance would still work.

claudep:
I created a basic test to see if pickling was somewhere broken. Either I didn't reproduce the exact conditions you had or the issue has been solved by some other recent commit. If you still care for this, could you check it?

woozyking:
Apologies as this has been a bit too far back to recall on the exact issue. I think it's fairly safe to assume this is no longer an issue as I also could not reproduce it now.

Closing.

PR #405, opened 2019-10-19
Refs #293 - Ensured Dataset can be pickled/unpickled without damages

claudep:


tests/test_tablib.py
```diff
@@ -460,6 +460,13 @@ def test_unicode_renders_markdown_table(self):
         self.founders.append(('Old', 'Man', 100500))
         self.assertEqual('first_name|last_name |gpa   ', unicode(self.founders).split('\n')[0])
 
+    def test_pickle_unpickle_dataset(self):
+        import pickle
```
hugovk: Let's keep the imports together at the top of the file.

claudep:
I suggest to wait a bit on reporter input before committing.

PR #408, opened 2019-10-19
Drop python 2 support

claudep:
Remove support python 2 from doc, requirements.txt and config
Replace unicode with str
Remove dbfpy folder and rename dbfpy3 to dbfpy
Remove compat file and remove python2 packages from dependency


claudep:
Just to confirm travis is OK...

PR #410, opened 2019-10-19
Remove Python 2 code

hugovk:
Some extra cleanup for https://github.com/jazzband/tablib/issues/390.

* Remove some `if sys.version_info < (2, 4):` code
* Run https://github.com/asottile/pyupgrade with `--py3-plus`
* Remove unnecessary `__getslice__`, [deprecated since Python 2.0](https://docs.python.org/2/reference/datamodel.html#object.__getslice__) and removed in 3

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
 #
 # Tablib documentation build configuration file, created by
 # sphinx-quickstart on Tue Oct  5 15:25:21 2010.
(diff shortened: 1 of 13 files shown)


claudep:
I don't care so much, but replacing `%` placeholder syntax by `format()` brings nothing AFAIK in our case (contrary to the Python 3.6+ f-string syntax which might help for readability). But if you like this, feel free to go ahead with this patch.

hugovk:
`format()` is the more modern way in Python, and it's easier to use the `pyupgrade` defaults. We can run it again when dropping Python 3.5 and it'll do some f-string conversion too.

---

Also added some more tests, and include tests in coverage: https://nedbatchelder.com/blog/201908/dont_omit_tests_from_coverage.html

PR #414, opened 2019-10-22
Add test from #202

### LINKED ISSUE #202: Tablib import Excel NA() values as 42
If you import an xls or xlsx file into a dataset using `import_set` or `import_sheet` and one of the cells contains a `NA()` error value then the value of the imported cell in the dataset is 42.

That's probably not what you want in any circumstance.

See http://stackoverflow.com/questions/4928629/xlrd-excel-script-converting-n-a-to-42, for example.

You can work around that by changing:

```python
row_vals = [c.value for c in row]
```

to

```python
row_vals = [xlrd.error_text_from_code[c.value] if c.ctype == xlrd.XL_CELL_ERROR else c.value for c in row]
```

hugovk:
Adds the test from #202, now passes on Python 3.5-3.8.

Closes #202.

### DIFF
--- a/tests/test_tablib.py
+++ b/tests/test_tablib.py
@@ -3,11 +3,13 @@
 
 import datetime
 import doctest
+import io
 import json
 import pickle
 import unittest
 from uuid import uuid4
 
+import openpyxl
 import tablib
 from MarkupPy import markup
 from tablib.core import Row, detect_format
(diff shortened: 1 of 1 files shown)


claudep:
The reported issue talks about handling of `NA()` values. How is this test related to that?

hugovk:
Good question! In any case, https://github.com/jazzband/tablib/issues/202#issuecomment-181284179 does say that test passed on 2.7 and failed on 3.4.

hugovk:
Replaced by https://github.com/jazzband/tablib/pull/415.

Issue #340, opened 2018-10-28
export "cli" format


nskalis:
Hi,

Would you consider an `export` format suited for cli purposes as well, like

1. the `simple` table format as described here https://pypi.org/project/tabulate/
2. the same format is used by all tools from HashiCorp, Consul, Vault, etc.

in other words, it is a familiar more or less format.

Example:
```
Path          Type         Description
----          ----         -----------
cubbyhole/    cubbyhole    per-token private secret storage
kv/           kv           n/a
secret/       kv           key/value secret storage
sys/          system       system endpoints used for control, policy and debugging
```

xdanielsb:
Hi @nskalis 

now I am taking care of this feature, sorry for the delay

claudep:
@xdanielsb You are welcome to work on this, but note that we won't commit anything before #256 is reviewed and pushed.

xdanielsb:
Hi @nskalis 

Here you can find a tutorial, whereas the pull request is approved you can use this not official feature in the branch feature-cli-format in the following way,

```python
students = tb.Dataset()
students.headers = ['first', 'last']
students.rpush(['Kenneth', 'Reitz'], tags=['male', 'technical'])
students.rpush(['Daniel', 'Dupont'], tags=['male', 'creative'])
students.rpush(['Bessie', 'Monke'], tags=['female', 'creative'])

print( students.export('cli',tablefmt="github") )
```
| first   | last   |
|---------|--------|
| Kenneth | Reitz  |
| Daniel  | Dupont |
| Bessie  | Monke  |

It's that simple, you can use all formats that [tabulate](https://pypi.org/project/tabulate/) offers

claudep:
The commit was reverted.

claudep:
This functionality was pushed in fa30ea858d0ac41f2f542e0c50d9f9a2e343deb2

PR #419, opened 2019-11-06
Implement feature CLI format as a wrapper of python-tabulate.

xdanielsb:
- This solve the issue #340
- Write test for 'plain'  and 'github' format

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -9,6 +9,7 @@ install = [
     'xlrd',
     'xlwt',
     'pyyaml',
+    'tabulate',
 ]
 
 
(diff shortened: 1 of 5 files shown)


claudep:
Generally speaking, I'm a bit worried that with each format we are adding, we are forcing users to install more code even if they never use some formats. I'd like to have more formats like `df` where the dependency is optional.
I will create a separate issue to discuss this, so maybe wait on that discussion before changing anything here.

claudep:
This was rewritten in #437 and merged.

PR #420, opened 2019-11-06
Moved format documentation from code to docs

claudep:


claudep:
@hugovk What do you think about this proposal patch? I prefer having an explicit help page rather than counting on automatic API, but it can be a matter of taste.

hugovk:
👍🏼 

Auto docs can often be a little lacking, so explicit docs are often better. Just need to remember to update them as the code changes.

docs/formats.rst
```diff
@@ -0,0 +1,184 @@
+.. _formats:
+
+=======
+Formats
+=======
+
+Tablib supports a wide variety of different tabular formats, both for input and
+output. Moreover, you can :ref:`register your own formats <newformats>`.
+
+You'll find below the formats that are shipped with Tablib.
```
hugovk: This sentence is redundant and can be removed.

docs/formats.rst
```diff
@@ -0,0 +1,184 @@
...
+===
+
+When you import CSV data, you can specify if the first line of your data source
+is headers with the ``headers`` boolean parameter (defaults to ``True``)::
+
+    import tablib
+
+    tablib.import_set(your_data_stream, format='csv', headers=False)
+
+When exporting with the ``csv`` format, the top row will contain  headers, if
```
hugovk: Remove second space before "headers"

PR #424, opened 2019-11-08
Point README to the documentation

claudep:


claudep:
@hugovk what about simplifying the README and not have to check documentation correctness at two different places?

hugovk:
Good idea!

PR #425, opened 2019-11-08
Fixes #421 - Make all dependencies optional

### LINKED ISSUE #421: Render more formats optional
When you install tablib, a rather big numbers of dependencies are automatically installed. It may not be a big deal on performant system, but still I think we should not force users to install dependencies for formats they may never use.

The downside of optional deps is that the install command is a bit more complicated, and requires the user to think a bit of desired formats. (e.g. `pip install tablib[xlsx, ods, yaml]`).
The formats that could be optional (= require dependencies) are: `xls`, `xlsx`, `odf`, `yaml`, `html` (and `dbf` when we'll succeed in externalizing it).

claudep:


bluetech:
*Thank you*! It will be great to get rid of all of these dependencies.

I wonder if this is compatible with how django-import-export [does things](https://github.com/django-import-export/django-import-export/blob/master/import_export/formats/base_formats.py)?

claudep:
I think there are two options:
 - if `django-import-export` wants to support all formats as before, it just has to change its dependencies to include the tablib optional ones.
 - if `django-import-export` wants to let the user choose the installed dependencies, it should do something similar to this patch, and then check with the tablib registry to know if some format is available or not.

I'm happy to help if needed.

hugovk:
Perhaps a `pip install tablib[all]` shortcut could install all?

hugovk:
And I think this would warrant a version bump.

claudep:
So is it valid/acceptable to add a line like `'all': ['odfpy', 'markuppy', 'pandas', 'pyyaml', 'xlrd', 'xlwt', 'openpyxl>=2.4.0'],` in `extras_require`? 

For the version bump, aren't we going to bump the version for the next release anyway?

hugovk:
> So is it valid/acceptable to add a line like `'all': ['odfpy', 'markuppy', 'pandas', 'pyyaml', 'xlrd', 'xlwt', 'openpyxl>=2.4.0'],` in `extras_require`?

Yes, something like that. compare what attrs does:

https://github.com/python-attrs/attrs/blob/179b1c4bf0fedaa0e0281649719d1f1e0bd51e94/setup.py#L39-L55

> For the version bump, aren't we going to bump the version for the next release anyway?

Yep :)

setup.py
```diff
@@ -42,8 +32,12 @@
...
     ],
     python_requires='>=3.5',
-    install_requires=install,
     extras_require={
+        'ods': ['odfpy'],
+        'html': ['markuppy'],
         'pandas': ['pandas'],
+        'yaml': ['pyyaml'],
+        'xls': ['xlrd', 'xlwt'],
+        'xlsx': ['openpyxl>=2.4.0'],
```
hugovk: Let's alphabetise these

tests/test_tablib.py
```diff
@@ -49,6 +50,16 @@ def _test_export_data_in_all_formats(self, dataset, exclude=()):
                 continue
             dataset.export(format_)
 
+    def test_unkown_format(self):
```
hugovk: Typo:
```suggestion
    def test_unknown_format(self):
```

setup.py
```diff
@@ -33,11 +33,12 @@
     ],
     python_requires='>=3.5',
     extras_require={
-        'ods': ['odfpy'],
+        'all': ['odfpy', 'markuppy', 'pandas', 'pyyaml', 'xlrd', 'xlwt', 'openpyxl>=2.4.0'],
```
hugovk: And let's alphabetise these too :)

claudep: eagle eye :)

docs/install.rst
```diff
@@ -19,8 +19,20 @@ Of course, the recommended way to install Tablib is with `pip <https://pip.pypa.
 
 .. code-block:: console
 
-    $ pip install tablib[pandas]
+    $ pip install tablib
 
+You can also choose to install more dependencies to have more import/export
```
hugovk: Should we also show an example of installing with a single dependencies?

claudep: Good idea, done.

claudep:
Merged in 626a062, thanks for the review!

PR #426, opened 2019-11-10
Update testing: add docs + lint jobs; use pre-commit for linting

hugovk:
* Travis CI: Move docs and lint to their own build job for more parallelism, this way we have the max 5 jobs running at once, instead of 3, and can quickly see if, say, the linting step failed

* GitHub Actions: Add a workflow to run the lint and docs tests

* tox: move isort into pre-commit. See instructions below about pre-commit; for tox, and in the CI, it's `pre-commit run all files`. For the user, they can still run `tox -e lint` as before.

---

# pre-commit

pre-commit can be used locally to check files when committing. If problems are found, it stops the commit with an error message. Some tools also fix the problems they find.

Fail fast, fail early.

pre-commit is not for everyone, and it's not mandatory to run it locally. I didn't used to be keen, but now I find it really helpful. It's important it doesn't take too long to run.

Even if you don't run it locally, it's handy for grouping together many linters, and running on the CI.

Here are the main commands for local use.

```bash
# Do this just once to init a repo:
pre-commit install

# The first run can take a bit longer to set up, after that it's quick

# Then commit like normal, it’ll run the checks on only the staged files:
git commit -m "something"

# If you want to skip the pre-commit checks:
git commit -m "something" -n

# If you want to run on all files:
pre-commit run all-files

# Once in a while, check for new versions of your lint tools:
pre-commit autoupdate  # and then commit the updated config file

# Stop using it locally:
pre-commit uninstall
```

I added some extra handy linters, such as `check-yaml`, which is really handy to find syntax errors in, say, `.travis.yml` before pushing and trying to find out why nothing was built.

We can also add Black and Flake8 here, when we're ready.

`rst-backticks` found some problems, which I've fixed, and `pyupgrade` made a couple of upgrades, which are included.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,6 +1,9 @@
 dist: xenial
 language: python
-cache: pip
+cache:
+  pip: true
+  directories:
+    - $HOME/.cache/pre-commit
 
 matrix:
   fast_finish: true
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,13 +1,24 @@
-language: python
-python:
-- 3.6
-- 3.7
-- 3.8
-cache: pip
 dist: xenial
+language: python
+cache: pip
+
+matrix:
+  fast_finish: true
+  include:
+    - python: 3.8
+      env: TOXENV=docs
+    - python: 3.8
+      env: TOXENV=lint
+    - python: 3.8
+    - python: 3.7
+    - python: 3.6
+
 install: travis_retry pip install tox-travis
+
 script: tox
+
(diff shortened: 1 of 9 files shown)


.travis.yml
```diff
@@ -1,13 +1,24 @@
+dist: xenial
```
claudep: If I'm not mistaken, this is now the default and could be removed.

hugovk: Yep, removed!

PR #427, opened 2019-11-10
Format with Black

hugovk:
For #401

Locally:

```bash
# Install
pip install -U black

# Run
black .
```

Also run as part of:

```bash
pre-commit run --all-files

# Or
pre-commit run --all-files black

# Or
tox -e lint
```

Also run as part of the CI lint jobs.

PR #428, opened 2019-11-10
Fix Flake8

### LINKED ISSUE #401: Consider enforcing flake8, black and isort
As part of #398 I've added disabled calls to flake8 including a commented-out installation of the flake8-black and flake8-isort plugins. I would suggest to run black and isort once over the code base and enabling both plugins to have a positive long-term effect on code quality. I didn't make this part of #398 since it's not a hard requirement for Jazzband projects (yet) but I've seen more and more projects move to that strategy.

hugovk:
Fixes #401.

Includes https://github.com/jazzband/tablib/pull/427, the first two commits. The others are new to this PR.

Flake8 found one bug: https://github.com/jazzband/tablib/commit/e3b445cec3ed25fc3c972fd3982efd81629f997a

And one bit of old Python 2 code: https://github.com/jazzband/tablib/commit/77c54885878dbbcbef1fb2aa68ed116f4f266ff1

Locally:

```bash
# Install
pip install -U flake8

# Run
flake8
```

Also run as part of:

```bash
pre-commit run --all-files

# Or
pre-commit run --all-files flake8

# Or
tox -e lint
```

Also run as part of the CI lint jobs.


src/tablib/formats/_rst.py
```diff
@@ -123,7 +123,7 @@ def export_set_as_simple_table(cls, dataset, column_widths=None):
         lines = []
         wrapper = TextWrapper()
         if column_widths is None:
-            column_widths = _get_column_widths(dataset, pad_len=2)
+            column_widths = cls._get_column_widths(dataset, pad_len=2)
```
claudep: Would be great to have a separate PR for this, with a test to complete coverage of that line.

hugovk: I think we could probably remove this whole `if` block. I don't see how `column_widths` can be `None`.

The default value is `None` in the method definition:

```python
    def export_set_as_simple_table(cls, dataset, column_widths=None):
```

But it's only called from one place in `export_set`, and it includes a variable for `column_widths`, so it won't default to `None`:

```python
        if use_simple_table and not force_grid:
            return cls.export_set_as_simple_table(dataset, column_widths)
        else:
            return cls.export_set_as_grid_table(dataset, column_widths)
```

Earlier in that method, it is set as:

```python
        column_widths = cls._get_column_widths(dataset, max_table_width)
```

Which will always return a list:

```python
        column_widths = [max(w, l) for w, l in zip(column_widths, word_lens)]
        return column_widths
```

A similar thing happens for the similar `if` block in `export_set_as_grid_table`.

These are all in `_rst.py`, the underscore denoting a private, internal code, and not part of the API. 

What do you think?

claudep: Not sure, as the method itself is not prefixed with underscore, so if someone could very well call something like `fmt = registry.get_format('rst');  fmt.export_set_as_simple_table(dataset)` without breaking private rules. No string opinion, but I would tend to fix and test.

hugovk: There we go! https://github.com/jazzband/tablib/pull/433

hugovk:
Let's not worry about Black for now, it's not essential.

PR #429, opened 2019-11-10
No __cmp__ or cmp in Python 3

hugovk:
Cherry-picked from #428, found by Flake8.

### DIFF
--- a/src/tablib/packages/dbfpy/fields.py
+++ b/src/tablib/packages/dbfpy/fields.py
@@ -101,6 +101,24 @@ class DbfFieldDef:
         self.start = start
         self.end = stop
 
+    def __eq__(self, other):
+        return repr(self) == repr(other)
+
+    def __ne__(self, other):
+        return repr(self) != repr(other)
+
+    def __lt__(self, other):
+        return repr(self) < repr(other)
+
+    def __le__(self, other):
+        return repr(self) <= repr(other)
+
+    def __gt__(self, other):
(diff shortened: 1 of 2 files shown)


claudep:
I wonder if we should replace this by rich comparison methods (https://portingguide.readthedocs.io/en/latest/comparisons.html#rich-comparisons).

hugovk:
That might be a good idea. Or do we have anywhere in code that is comparing these fields? How might a user do so, and how is it useful?

claudep:
I must admit I'm not more advanced on this question. It's just keeping the functionality without much thinking...

hugovk:
I've updated the PR to add rich comparisons, with tests.

I did it by comparing the `repr(self)` with `repr(other)`.

Previously it was comparing `self.name` with `str(other).upper()` which are different ways of looking at the data.

For example, these two fields would be different:

```python
a = fields.DbfCharacterFieldDef("abc", self.length)
b = fields.DbfCharacterFieldDef("abc", self.length)
```

Because `self.name` of `a` is `ABC` and `str(other).upper()` of `b` is `ABC        C  10   0`.

claudep:
That's great! Do you think you could simplify a bit by using the `total_ordering` decorator?

hugovk:
Done!

hugovk:
Done!

image

claudep:
Perfect, thanks.

PR #430, opened 2019-11-11
Fixes #422 - Allow ability to lazy-load external modules

### LINKED ISSUE #422: Load formats "on-demand"
When  you `import tablib` in a not-too-recent system, you have to wait a bit (~5 seconds on my system), mainly due to all libraries used by all tablib formats being loaded at startup even if they are never used.

I'd suggest to only load the format files/libraries when the format is required by an import or export.

claudep:


hugovk:
Looks good, a noticeable 2.411s -> 0.709s improvement for me (macOS Mojave, Python 3.7):

```
⌂73% [hugo:~/github/tablib] master* ± cat 1.py
import tablib
⌂69% [hugo:~/github/tablib] master* ± p 1.py
⌂68% [hugo:~/github/tablib] master* 2s ± time p 1.py
python3 1.py  1.42s user 0.36s system 73% cpu 2.411 total
⌂66% [hugo:~/github/tablib] master* 2s ± gco claudep/lazy_load
Note: checking out 'claudep/lazy_load'.

You are in 'detached HEAD' state. You can look around, make experimental
changes and commit them, and you can discard any commits you make in this
state without impacting any branches by performing another checkout.

If you want to create a new branch to retain commits you create, you may
do so (now or later) by using -b with the checkout command again. Example:

  git checkout -b <new-branch-name>

HEAD is now at 2744c38 Fixes #422 - Allow ability to lazy-load external modules
⌂65% [hugo:~/github/tablib] 2744c38* ± time p 1.py
python3 1.py  0.43s user 0.13s system 79% cpu 0.709 total
```

tests/test_tablib.py
```diff
@@ -306,6 +307,9 @@ def test_auto_format_detect(self):
         """Test auto format detection."""
         # html, jira, latex, rst are export only.
 
+        # Looks like openpyxl is indirectly causing a distutils deprecation warning
+        warnings.filterwarnings('ignore', module='distutils')
```
hugovk: Is this already reported upstream? If not let's do so, and perhaps include a link here so we can remove it one day.

claudep: According to https://github.com/pypa/virtualenv/blob/master/docs/changes.rst, this was fixed in virtualenv 16.4.0 released in February 2019. If this only annoys me as I'm using an older version, I'm fine removing that warning filter. Otherwise I can add a note about that.

hugovk: Are you able to upgrade to a newer virtualenv? If not, let's add a note and we can remove it when you upgrade.

claudep: My personal case should not be taken into account here. Having a warning is not an issue. I think I'll simply remove that part.

hugovk: Just tested, and I get another warning from openpyxl when running pytest:

```
================================================= warnings summary ==================================================
/usr/local/lib/python3.7/site-packages/openpyxl/xml/functions.py:30
  /usr/local/lib/python3.7/site-packages/openpyxl/xml/functions.py:30: DeprecationWarning: defusedxml.lxml is no longer supported and will be removed in a future release.
    from defusedxml.lxml import fromstring as _fromstring, tostring
```

I have latest `openpyxl==3.0.0` and also latest `defusedxml==0.6.0`, but guess this doesn't really matter if openpyxl stops using defusedxml and uses another one instead.

claudep: Seems to be a `defusedxml` deprecation (read https://github.com/tiran/defusedxml/issues/15). What would you suggest to do at our level?

hugovk: We're not relying on defusedxml.lxml, and don't install it. It's an optional dependency of openpyxl.

I've created an upstream PR to silence the warning:
* https://bitbucket.org/openpyxl/openpyxl/pull-requests/353/ignore-deprecation-warning-from/diff

We could also silence it, but let's see how this goes.

claudep: I removed the silencing. In a sense we cannot be responsible for warnings raised by dependencies, new ones can come at any time.

PR #437, opened 2019-12-04
Implement feature that allows to export tabular data suited to a terminal. This solves issue #340.

xdanielsb:
Hi I implemented a wrapper that helps to export tabular data into a Command Line Interface. All test with different versions of  formats were performed. I appreciate any comment or suggestion. If there is any additional  change or comment, please let me know.

PR #438, opened 2019-12-04
Refactor error raising to remove duplication

hugovk:


hugovk:
This will conflict with https://github.com/jazzband/tablib/pull/437. Shall we merge this before or after that one?

claudep:
Depends on whether you want to be nice to Daniel :-)

src/tablib/formats/__init__.py
```diff
@@ -134,7 +116,13 @@ def formats(self):
     def get_format(self, key):
         if key not in self._formats:
             if key in uninstalled_format_messages:
-                raise UnsupportedFormat(uninstalled_format_messages[key])
+                uninstalled_format_messages[key].update(key=key)
```
bluetech: Instead of modifying a global, which is not so nice, I'd suggest to add `key=key` to the `format(...)`.

hugovk: Thanks, updated in 2b364fcc20c7d07ebdbe0b932e7d2fbbcb3eb860.

hugovk:
Rebased!

claudep:
Could you just squash the first two commits (letting only the pre-commit stuff separate)?

hugovk:
OK, I can do that later. But won't "Squash and merge" nullify it in any case?

image

claudep:
Sure we could do that, but as the pyupgrade commit is not related to the main issue, I would have let it as a separate commit.

hugovk:
OK to squash and merge instead?

claudep:
Done :-) Thanks for the patch.

PR #439, opened 2019-12-11
Remove unnecessary MANIFEST.in

hugovk:
When using setuptools_scm, MANIFEST.in is needed when we want to exclude tracked files, or include untracked files.

https://github.com/pypa/setuptools_scm/blob/master/README.rst#file-finders-hook-makes-most-of-manifestin-unnecessary

It doesn't look like we need those, and check-manifest causes the CI to sometimes fail when adding new files, which setuptools_scm will find anyway. 

Ping @jezdez

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,6 +1,0 @@
-recursive-include docs *
-recursive-include tests *
-include pyproject.toml pytest.ini tox.ini .coveragerc .pre-commit-config.yaml *.md LICENSE AUTHORS
-prune docs/_build
-prune *.pyc
-prune __pycache__
--- a/tox.ini
+++ b/tox.ini
@@ -17,7 +17,6 @@ commands =
 
 [testenv:lint]
 deps =
-    check-manifest
     flake8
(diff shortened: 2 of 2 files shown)


jezdez:
Ah, huh, that makes total sense, I forgot that setuptools_scm does that. Would you mind -- just for the record -- create a sdist and bdist_wheel with and without the manifest template and compare the file contents?

hugovk:
Here's the sdist and wheels for master and this PR:

* [dist.zip](https://github.com/jazzband/tablib/files/3949585/dist.zip)

Only expected changes in the sdist

image

image

image

And only expected changes in the wheel (renamed here to `.zip` for ease of comparison):

image

image

image

image

jezdez:
You're the best, thanks @hugovk!

hugovk:
You're welcome! Thanks to @dyve who brought up the issue for Black:

* https://github.com/psf/black/issues/1199

jezdez:
I think there is a chance that other Jazzband project have done the same 😬

PR #448, opened 2020-01-25
Optimize xlsx detection

claudep:
Reading the whole file is a bit too much to detect if the file
looks like an xlsx file.

### DIFF
--- a/src/tablib/formats/_xlsx.py
+++ b/src/tablib/formats/_xlsx.py
@@ -3,12 +3,12 @@
 
 from io import BytesIO
 
-import openpyxl
 import tablib
-
-Workbook = openpyxl.workbook.Workbook
-ExcelWriter = openpyxl.writer.excel.ExcelWriter
-get_column_letter = openpyxl.utils.get_column_letter
+from openpyxl.reader.excel import ExcelReader, load_workbook
+from openpyxl.styles import Alignment, Font
+from openpyxl.utils import get_column_letter
+from openpyxl.workbook import Workbook
+from openpyxl.writer.excel import ExcelWriter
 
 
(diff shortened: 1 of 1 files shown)


hugovk:
Sounds reasonable.

I don't suppose this is testable directly? It should successfully load a file before the change, and it should load a file afterwards too... Would be good at least to check that the changed lines are covered by tests (and https://github.com/jazzband/tablib/pull/449 should fix coverage reports in PRs).

src/tablib/formats/_xlsx.py
```diff
@@ -3,12 +3,12 @@
...
 import tablib
-
-Workbook = openpyxl.workbook.Workbook
-ExcelWriter = openpyxl.writer.excel.ExcelWriter
-get_column_letter = openpyxl.utils.get_column_letter
+from openpyxl.reader.excel import ExcelReader, load_workbook
+from openpyxl.styles import Alignment, Font
+from openpyxl.utils import get_column_letter
+from openpyxl.workbook import Workbook
+from openpyxl.writer.excel import ExcelWriter
```
hugovk: I don't mind these changes, although it does take away some context when they're used below.

claudep: I understand your remark, but as the entire file is about xlsx as its name implies, I don't think that this "path" context is so useful here.

hugovk: Good point!

claudep:
I think all detection code is tested in `test_auto_format_detect` (and there's even a specific `test_xlsx_format_detect`). I'll rebase and force push to see if there's now a coverage diff report.

PR #454, opened 2020-02-07
Fixes #453 - Reversing behavior of Row.lpush/Row.rpush

### LINKED ISSUE #453: Row.lpush/rpush behavior is reversed
As discovered while testing #450, `Row.lpush` is appending, while `Row.rpush` is prepending, which is counter-intuitive (and doesn't match the `Dataset.lpush/rpush`behavior).

claudep:
Co-authored-by: chim <chenpan@xiaomai5.com>

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,13 @@
 # History
 
+## 2.0.0 (Unreleased)
+
+### Breaking changes
+
+- The `Row.lpush/rpush` logic was reversed. `lpush` was appending while `rpush`
+  and `append` were prepending. This was fixed (reversed behavior). If you
+  counted on the broken behavior, please update your code (#453).
+
 ## 1.1.0 (2020-02-13)
 
 ### Deprecations
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -71,10 +71,10 @@ class Row:
             setattr(self, k, v)
 
     def rpush(self, value):
-        self.insert(0, value)
+        self.insert(len(self._row), value)
 
     def lpush(self, value):
-        self.insert(len(value), value)
+        self.insert(0, value)
 
     def append(self, value):
         self.rpush(value)
--- a/tests/test_tablib.py
+++ b/tests/test_tablib.py
@@ -556,27 +556,15 @@ class TablibTestCase(BaseTestCase):
 
     def test_row_lpush(self):
         """Row lpush."""
-        # Arrange
         john = Row(self.john)
-        george = Row(self.george)
-
-        # Act
-        john.lpush(george)
-
-        # Assert
-        self.assertEqual(john[-1], george)
+        john.lpush(53)
+        self.assertEqual(john.list, [53, 'John', 'Adams', 90])
 
     def test_row_append(self):
         """Row append."""
-        # Arrange
         john = Row(self.john)
-        george = Row(self.george)
-
-        # Act
-        john.append(george)
-
-        # Assert
-        self.assertEqual(john[0], george)
+        john.append('stuff')
+        self.assertEqual(john.list, ['John', 'Adams', 90, 'stuff'])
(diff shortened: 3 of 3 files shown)


hugovk:
HISTORY.md correctly notes this as a breaking change, and it's definitely fixing a bug. 

However, we of course don't know how many people are using this and relying on the broken behaviour. 

The cautious thing to do would be to use [SemVer](https://semver.org/) and put (deprecation?) warnings in release 1.10, and then make the breaking change in release 2.0.

claudep:
Providing a deprecation period is nice when the user can do something about it in its code, like using another path to avoid the warning. Here, what would we suggest to do to avoid the deprecation in 1.1.0?

hugovk:
Some options:

1. Just fix it, release as breaking change in 1.1.
   * Not following SemVer
   * May break code where people are relying on current behaviour
   * Non-major bump may surprise people who expect no breaking changes

2. Just fix it, release as breaking change in 2.0.
   * Follows SemVer
   * May break code where people are relying on current behaviour
   * But a major bump indicates breaking changes, more likely to check release notes

3. Deprecation warning in 1.1. Fix in 2.0.
   * Users can fix it and silence the warning
   * But silencing specific warnings isn't always easy, especially if it comes from a dependency
   * Not always straightforward to silence warnings
   * Need to un-silence for 2.0?

4. Introduce new, fixed methods in 1.1. Deprecate old, broken ones in 1.1. Remove old, broken ones in 2.0.
   * Warns people immediately about the bug
   * They can start using the fixed ones
   * No particular rush to get 2.0 out

I favour number 4. 

For example, the new ones could be `l_push`/`r_push` or `left_push`/`right_push` or `prepend`/`append`.

Or better, we should look to the stdlib and use consistent naming with existing data structures.

* [Lists](https://stackoverflow.com/q/8537916/724176) have `l.append(x)`, and `l.insert(0, x)` (there's no `prepend` because it's [inefficient](https://mail.python.org/pipermail/tutor/2005-March/036803.html).

* [`collections.deque`](https://docs.python.org/3/library/collections.html#collections.deque) has `append` and `appendleft`.

How about this?

hugovk:
Hmm, a slight complication. The existing `append` just calls `rpush` and is therefore also now broken (and is fixed by this PR, so needs to be in the release notes):

```python
    def append(self, value):
        self.rpush(value)
```

claudep:
My feeling is that few people are using this `Row` API. And if they are, they should have noticed that reversed behavior. However, I didn't read any bug report about that, so at some point, if they didn't report it, then they are also partly to blame. That's why I still prefer 1. or 2. A drawback of changing the API is that it will be different from the `Dataset` object which also has `lpush` and `rpush`.

hugovk:
Good point, `Row` also has a similar `lpush_col` and `rpush_col`.

How about we go for option 2?

And perhaps make a 1.1 release first where we can at least mention the upcoming change in the release notes?

HISTORY.md
```diff
@@ -1,5 +1,13 @@
 # History
 
+## 1.1.0 (Unreleased)
+
+### Breaking changes
+
+- The `Row.lpush/rpush` logic was reversed. `lpush` was appending while `rpush`
+  was prepending. This was fixed (reversed behavior). If you counted on the
```
hugovk:
```suggestion
  and `append` were prepending. This was fixed (reversed behavior). If you counted on the
```

claudep:
That sounds like a good plan.

HISTORY.md
```diff
@@ -1,5 +1,13 @@
 # History
 
+## 1.1.0 (Unreleased)
```
hugovk:
```suggestion
## 2.0.0 (Unreleased)
```

And something like https://github.com/jazzband/tablib/pull/455 for 1.1?

PR #461, opened 2020-02-25
Update xlsx import to read cell values instead of cell formulas

dragonworks:
When importing an Excel xlsx file it reads the cell formulas instead of the cell values.

For example, if a cell has the formulas '=SUM(A1:A3)' it will read that string instead of the calculated number in Excel.

Passing data_only=True to openpyxl's load_workbook will load the cell values instead of formulas.

I think when importing an Excel spreadsheet it's the values that would be wanted and not the formulas?


claudep:
I think that makes sense. Could you please add:
 * a Changelog entry in HISTORY.md
 * a note in `docs/formats.rst` under the `xlsx` header
 * a test reading an xlsx file containing a formula

dragonworks:
Thanks, I've updated it with those changes. Hope that's okay, I'm kinda new to this.

claudep:
That's great, thanks! I pushed minor edits to your patch, if you don't mind.

dragonworks:
Thanks, I appreciate it!

docs/formats.rst
```diff
@@ -206,6 +206,15 @@ Import/export data in Excel 07+ Spreadsheet representation.
...
 make the format available.
 
+.. note::
+
+    When reading an ``xlsx`` file containing formulas in its cells, Tablib will
+    read the cell values, not the cell formulas.
+
+.. versionchanged:: 2.0.0
+
+    Reading now cell values instead of formulas.
```
hugovk: For brevity: 
```suggestion
    Reads cell values instead of formulas.
```

claudep: Sure!

PR #475, opened 2020-09-20
fix database.export(json),yaml in chinese will get u787979 problems

Galaxyjia:
when load *.xlsx in 'rb' and have some chinese in data, it will keep unicode in data.  such as '\u8979766'
and when try to export in json or yaml,it will still keep the unicode in format

```python
data = tablib.Dataset()
with open('data.xlsx','rb') as fh:
    data.load(fh,'xlsx')
print(data.dict)
print(json.dumps(data.dict,ensure_ascii=False))

with open('data.json','w') as f:
    f.write(data.export('json'))

with open('data.json','w') as f:
    f.write(json.dumps(data.dict,ensure_ascii=False))
```

### DIFF
--- a/src/tablib/formats/_json.py
+++ b/src/tablib/formats/_json.py
@@ -23,12 +23,12 @@ class JSONFormat:
     @classmethod
     def export_set(cls, dataset):
         """Returns JSON representation of Dataset."""
-        return json.dumps(dataset.dict, default=serialize_objects_handler)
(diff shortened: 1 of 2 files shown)


hugovk:
Please could you add tests? This will help ensure there are no future regressions.

Galaxyjia:
> Please could you add tests? This will help ensure there are no future regressions.

ok,I will add tests in a few days.

claudep:
I completed this patch in #484.

PR #483, opened 2020-12-02
Expose 'read_only' parameter for 'import_set' and 'import_book'

### LINKED ISSUE #482: Loading xlsx in read_only mode yields giant sheet
Loading spreadsheet [tst.xlsx](https://github.com/jazzband/tablib/files/5632112/tst.xlsx) with `read_only=True` results in tablib trying to construct a sheet with 1048576 rows, which can take a long time and/or exhaust available memory:
```python
from openpyxl import load_workbook
wb=load_workbook(filename='tst.xlsx', read_only=True)
wb.active.max_row
1048576
```
with `read_only=False`, openpyxl returns the expected number of rows:
```python
from openpyxl import load_workbook
wb=load_workbook(filename='tst.xlsx', read_only=False)
wb.active.max_row
4
```
Openpyxl has this comment about read-only mode:

> _Read-only mode relies on applications and libraries that created the file providing correct information about the worksheets, specifically the used part of it, known as the dimensions. Some applications set this incorrectly._

so it seems to me it'd be safer if tablib were to use `read_only=False` or are there other issues with doing that?  If the latter is the case, would it be possible to provide a means to force a spreadsheet to be loaded with read-only turned off?  Unfortunately, we don't have any control over the tool that's generating the "incorrect" files as they come from a third-party web site.

[fix.diff.txt](https://github.com/jazzband/tablib/files/5632169/fix.diff.txt)

dmosberger:
Here is a proposed fix for issue #482.

Fixes #482.


claudep:
Thanks, it looks good to me! Regarding docs, I would add an entry in `HISTORY.md` and document the parameter in `formats.rst`?

claudep (approved):
Thanks a lot!

hugovk (approved):

hugovk:
Thank you!

PR #485, opened 2020-12-05
Bumped version to 3.0.0

claudep:


claudep:
@hugovk What about releasing today? Any other short-term plan?

hugovk:
Yes, I think we're ready for a release!

We've dropped Python 3.5. Are we using semver and does this qualify for a major bump?

claudep:
I don't care so much, but if you think we should rather release 3.0.0, I'm fine with that.

HISTORY.md
```diff
@@ -1,6 +1,6 @@
 # History
 
-## Unreleased
+## 2.1.0 (2020-12-05)
```
hugovk, line 3: Okay, let's bump to 3.0.0, might save some surprises for people.

```suggestion
## 3.0.0 (2020-12-05)
```

Would you like me to do the release or would you like to?

claudep: If you have time for it, go!

hugovk: 🚀

PR #487, opened 2020-12-09
Better exceptions

nuno-andre:
It's best practise for libraries to have a common parent class for its custom exceptions so that they all can be easily catched.

This PR also refines the built-in exceptions inheritance and adds autodoc for the exceptions module.

### DIFF
--- a/docs/api.rst
+++ b/docs/api.rst
@@ -30,7 +30,6 @@ Databook Object
    :inherited-members:
 
 
-
 ---------
 Functions
 ---------
@@ -46,19 +45,8 @@ Exceptions
 ----------
 
 
-.. class:: InvalidDatasetType
-
-    You're trying to add something that doesn't quite look right.
-
-
-.. class:: InvalidDimensions
-
-    You're trying to add something that doesn't quite fit right.
-
-
-.. class:: UnsupportedFormat
-
-    You're trying to add something that doesn't quite taste right.
+.. automodule:: tablib.exceptions
+   :members:
 
 
 Now, go start some :ref:`Tablib Development <development>`.
--- a/src/tablib/exceptions.py
+++ b/src/tablib/exceptions.py
@@ -3,11 +3,11 @@ class TablibException(Exception):
 
 
 class InvalidDatasetType(TablibException, TypeError):
-    """Only Datasets can be added to a DataBook."""
+    """Only Datasets can be added to a Databook."""
 
 
 class InvalidDimensions(TablibException, ValueError):
-    """Invalid size."""
+    """The size of the column or row doesn't fit the table dimensions."""
 
 
 class InvalidDatasetIndex(TablibException, IndexError):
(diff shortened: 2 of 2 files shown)


hugovk:
Will close and re-open to see if we get an auto docs build!

(https://github.com/jazzband-roadies/help/issues/211)

hugovk:
It worked! Docs render properly:

https://tablib--487.org.readthedocs.build/en/487/api/#module-tablib.exceptions

Do we want to use the old exception doc descriptions as exception text, or use these new ones?

claudep:
The new descriptions are better, IMHO, but they could be even better. E.g. `Invalid size` is too generic, we could explain a little bit more.

nuno-andre:
Actually not new, I've only changed the strings to docstrings ;)

Maybe something like _"You're trying to add something that doesn't quite look right: Only Datasets can be added to a DataBook"_?

There were some missing exceptions in docs:

| exception | docstrings | docs |
| --- | --- | --- |
| `InvalidDatasetType` | _Only Datasets can be added to a DataBook._ | _You're trying to add something that doesn't quite look right._ |
| `InvalidDimensions` | _Invalid size._ | _You're trying to add something that doesn't quite fit right._ |
| `InvalidDatasetIndex` | _Outside of Dataset size._ |  |
| `HeadersNeeded` | _Header parameter must be given when appending a column in this Dataset._ | |
| `UnsupportedFormat` | _Format is not supported._ | _You're trying to add something that doesn't quite taste right._ |
| `TablibException` | _Tablib common exception._ |   |

nuno-andre:
@claudep I totally agree. Do you mean in code, where they're raised, or adding the message to exception? Something like:

```
class InvalidDimension(TablibException):
    '''A more specific message.
    '''
    def __init__(self, msg=None, *args, **kwargs):
        if not msg:
            msg = self.__doc__
        super().__init__(msg, *args, **kwargs)
```

claudep:
No, I was only referring to the docstring. In the code, the message should be the best possible in the place where it is raised.

nuno-andre:
E.g. _The size of the column or row doesn't fit the table dimensions._?

Also, do you think it would be worth adding the previous snippet to `TablibException` so that a custom exception has an explicit message about the error if it's raised without it?

claudep:
I'm not sure, if you find in the code examples where either the exception name or the exception message is not clear enough, then let's fix that in the code.

nuno-andre:
Of all the custom exceptions, only some `UnsupportedFormat` are raised with a message. In addition, there are some built-ins that I think could be reraised as `tablib` exceptions (with proper error messages). It would probably be necessary to add some other custom execeptions. 

A base exception is always a plus when adding a library as a dependency. Happy to contribute to this revision.

PS: btw, I just found that [`pandas` is out of the `tablib`'s import machinery](https://github.com/jazzband/tablib/blob/master/src/tablib/formats/_df.py#L4-L7).

claudep (approved):
Thanks, looks good.

PR #490, opened 2021-02-28
Fix "Invalid character / found in sheet title" during export to XLSX

### LINKED ISSUE #489: "Invalid character / found in sheet title" during export to XLSX
Hi there, 

it looks like tablib could automatically replace some characters when exporting to XLS, what do you think? The problem occurs when I export a sheet with a slash inside its title:

```
  File "/Users/mpasternak/.virtualenvs/bpp/lib/python3.8/site-packages/tablib/formats/_xlsx.py", line 53, in export_book
    ws.title = dset.title if dset.title else 'Sheet%s' % (i)
  File "/Users/mpasternak/.virtualenvs/bpp/lib/python3.8/site-packages/openpyxl/workbook/child.py", line 93, in title
    raise ValueError(msg)
ValueError: Invalid character / found in sheet title
```

Devs, please let me know if this is desired and I'll think about a patch.

mpasternak:
Fixes #489.

Make sure there are no bad characters in sheet names!

### DIFF
--- a/src/tablib/formats/_xlsx.py
+++ b/src/tablib/formats/_xlsx.py
@@ -2,6 +2,7 @@
 """
 
 from io import BytesIO
+import re
 
 from openpyxl.reader.excel import ExcelReader, load_workbook
 from openpyxl.styles import Alignment, Font
@@ -10,6 +11,11 @@ from openpyxl.workbook import Workbook
 from openpyxl.writer.excel import ExcelWriter
 
 import tablib
+
+INVALID_TITLE_REGEX = re.compile(r'[\\*?:/\[\]]')
+
+def safe_xlsx_sheet_title(s, replace="-"):
+    return re.sub(INVALID_TITLE_REGEX, s, replace)[:31]
 
 
 class XLSXFormat:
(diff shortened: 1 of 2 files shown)


src/tablib/formats/_xlsx.py
```diff
@@ -52,6 +67,8 @@ def export_book(cls, databook, freeze_panes=True):
             ws = wb.create_sheet()
             ws.title = dset.title if dset.title else 'Sheet%s' % (i)
 
+            #
+
```
claudep, line 71: I guess those two lines should not be in the patch.

claudep (changes requested):
Could you please include a small description in the commit first line, after the issue number?

src/tablib/formats/_xlsx.py
mpasternak: Updated, thanks

mpasternak:
> Could you please include a small description in the commit first line, after the issue number?

I am not sure if I understand, what you mean. 

Do you want me to change the commit message? Is this even possible? If it is, I have no idea how can I do that. 

I could re-write everything (as this is a very small update) and re-commit with a better message, but is that what you want from me? 

Please explain. I am willing to work with you on this patch but I am not sure if I understood correctly.

claudep:
Exactly, I'd like a small description of the change in the commit message.

You can do that by:
`git commit --amend` (and editing the message)
Then force push your branch:
`git push -f <repo-alias> master`

claudep:
Oh sorry, now I realize that you have several commits already. So you should use some rebase magic to do that.

`git rebase -i HEAD~3` would allow you to re-edit the commit message. I'd suggest you to study a bit the rebase capabilities, as it could surely be useful in your projects. Of course, rebasing should be done with caution and only on "working" branches.

mpasternak:
I did rebase and a force-push. Does it look okay for you?

claudep (approved):
Perfect, thanks a lot!

PR #509, opened 2021-10-30
Upgrade Sphinx used by Read the Docs

hugovk:
The docs build failure in https://github.com/jazzband/tablib/pull/508 / https://readthedocs.org/projects/tablib/builds/15132767/ is likely due to using an old version of Sphinx (1.8.5; Mar 10, 2019) which isn't compatible with the new docutils 0.18 release.

The newest Sphinx (4.2.0; Sep 12, 2021) isn't either, but it pins to a compatible docutils.

See https://github.com/sphinx-doc/sphinx/issues/9727.

~Creating a draft PR: we might need to iterate in case we need to specify extra requirements.~ Ready!

* https://tablib--509.org.readthedocs.build/en/509/


claudep (approved):
Thanks!

Issue #318, opened 2017-12-13
How does the cell width adapt


geekwolf:
How does the cell width adapt？

PR #516, opened 2022-02-17
feat: support column_width in xlsx format

Birdi7:
This PR allows user to specify column width in XLXS format. For it has a default value, which is not suitable for long values in dataset. As a result, it is not convenient to look these tables. Also, it eliminates questions like #318 in future

I didn't find tests for exporting, so omitted this. Let me know if they are required

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -34,3 +34,4 @@ Here is a list of past and present much-appreciated contributors:
     Tsuyoshi Hombashi
     Tushar Makkar
     Yunis Yilmaz
+    Egor Osokin
--- a/AUTHORS
+++ b/AUTHORS
@@ -9,6 +9,7 @@ Here is a list of past and present much-appreciated contributors:
     Bruno Soares
     Claude Paroz
     Daniel Santos
+    Egor Osokin
     Erik Youngren
     Hugo van Kemenade
     Iuri de Silvio
--- a/AUTHORS
+++ b/AUTHORS
@@ -4,6 +4,7 @@ by the Jazzband GitHub team.
 Here is a list of past and present much-appreciated contributors:
 
     Alex Gaynor
+    Andrew Graham-Yooll
     Andrii Soldatenko
     Benjamin Wohlwend
     Bruno Soares
--- a/AUTHORS
+++ b/AUTHORS
@@ -35,4 +35,3 @@ Here is a list of past and present much-appreciated contributors:
     Tsuyoshi Hombashi
     Tushar Makkar
     Yunis Yilmaz
-    Egor Osokin
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -251,9 +251,14 @@ can set to a number of lines that should be skipped before starting to read
 data.
 
 The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
-value you pass, the column width will be set accordingly. It can be either None, int, or "adapt".
-If "adapt" is passed, the column width will be unique for every column and will be 
-calculated based on values' length
+value you pass, the column width will be set accordingly. It can be either None, int, or "adaptive".
+If "adaptive" is passed, the column width will be unique for every column and will be 
+calculated based on values' length. Example of usage
+
+```python3
+data = tablib.Dataset()
+data.export('xlsx', column_width='adaptive')
+```
 
 
 .. versionchanged:: 3.3.0
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -261,7 +261,7 @@ For example::
 
 .. versionchanged:: 3.8.0
     The ``column_width`` parameter for ``export_set()`` was added.
-    
+
 .. versionchanged:: 3.1.0
 
     The ``skip_lines`` parameter for ``import_set()`` was added.
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -251,8 +251,8 @@ can set to a number of lines that should be skipped before starting to read
 data.
 
 The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
-value you pass, the column width will be set accordingly. It can be either ``None``, an integer, or "adaptive".
-If "adaptive" is passed, the column width will be unique for every column and will be 
+value passed, the column width will be set accordingly. It can be either ``None``, an integer, or default "adaptive".
+If "adaptive" is passed, the column width will be unique and will be 
 calculated based on values' length. For example:: 
 
     data = tablib.Dataset()
@@ -260,7 +260,7 @@ calculated based on values' length. For example::
 
 
 
-.. versionchanged:: 3.3.0
+.. versionchanged:: 3.8.0
     The ``column_width`` parameter for ``export_set()`` was added.
     
 .. versionchanged:: 3.1.0
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -250,6 +250,15 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
+value you pass, the column width will be set accordingly. It can be either None, int, or "adapt".
+If "adapt" is passed, the column width will be unique for every column and will be 
+calculated based on values' length
+
+
+.. versionchanged:: 3.2.0
+    The ``column_width`` parameter for ``export_set()`` was added.
+    
 .. versionchanged:: 3.1.0
 
     The ``skip_lines`` parameter for ``import_set()`` was added.
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -256,7 +256,7 @@ If "adapt" is passed, the column width will be unique for every column and will 
 calculated based on values' length
 
 
-.. versionchanged:: 3.2.0
+.. versionchanged:: 3.3.0
     The ``column_width`` parameter for ``export_set()`` was added.
     
 .. versionchanged:: 3.1.0
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -250,11 +250,11 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
-The ``export_set()`` method supports a ``column_width`` parameter. Depending 
-on the value passed, the column width will be set accordingly. It can be 
-either ``None``, an integer, or default "adaptive". If "adaptive" is passed, 
-the column width will be unique and will be calculated based on values' length. 
-For example:: 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending
+on the value passed, the column width will be set accordingly. It can be
+either ``None``, an integer, or default "adaptive". If "adaptive" is passed,
+the column width will be unique and will be calculated based on values' length.
+For example::
 
     data = tablib.Dataset()
     data.export('xlsx', column_width='adaptive')
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -250,15 +250,14 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
-The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
-value passed, the column width will be set accordingly. It can be either ``None``, an integer, or default "adaptive".
-If "adaptive" is passed, the column width will be unique and will be 
-calculated based on values' length. For example:: 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending 
+on the value passed, the column width will be set accordingly. It can be 
+either ``None``, an integer, or default "adaptive". If "adaptive" is passed, 
+the column width will be unique and will be calculated based on values' length. 
+For example:: 
 
     data = tablib.Dataset()
     data.export('xlsx', column_width='adaptive')
-
-
 
 .. versionchanged:: 3.8.0
     The ``column_width`` parameter for ``export_set()`` was added.
--- a/docs/formats.rst
+++ b/docs/formats.rst
@@ -251,14 +251,13 @@ can set to a number of lines that should be skipped before starting to read
 data.
 
 The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
-value you pass, the column width will be set accordingly. It can be either None, int, or "adaptive".
+value you pass, the column width will be set accordingly. It can be either ``None``, an integer, or "adaptive".
 If "adaptive" is passed, the column width will be unique for every column and will be 
-calculated based on values' length. Example of usage
+calculated based on values' length. For example:: 
 
-```python3
-data = tablib.Dataset()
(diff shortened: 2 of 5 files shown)


Birdi7:
Hello @hugovk! Sorry for ping, but I just want to ask if you have any time for review and merge this?

hugovk (changes requested):
Thanks for the ping! Please could you add some tests that use the different possible values?

Birdi7:
@hugovk Thank you for the review! I've added the test on check the new feature, checkout `test_xlsx_column_width`. Would you mind review the PR, and if everything is ok, merge it to master?

AUTHORS
```diff
@@ -32,3 +32,4 @@ Here is a list of passed and present much-appreciated contributors:
     Tommy Anthony
     Tsuyoshi Hombashi
     Tushar Makkar
+    Egor Osokin
```
hugovk, line 35: Let's keep alphabetical order.

docs/formats.rst
```diff
@@ -234,6 +234,20 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
+value you pass, the column width will be set accordingly. It can be either None, int, or "adaptive".
```
hugovk, line 238:
```suggestion
value you pass, the column width will be set accordingly. It can be either ``None``, an integer, or "adaptive".
```

docs/formats.rst
```diff
@@ -234,6 +234,20 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
...
+If "adaptive" is passed, the column width will be unique for every column and will be 
+calculated based on values' length. Example of usage
+
+```python3
+data = tablib.Dataset()
+data.export('xlsx', column_width='adaptive')
+```
```
hugovk, line 245: This is RST not Markdown, so we can use a double colon at the end of the last line to activate the code block:

```suggestion
calculated based on values' length. For example::

    data = tablib.Dataset()
    data.export("xlsx", column_width="adaptive")

File: src/tablib/formats/_xlsx.py [hunk 4]
```diff
@@ -35,12 +36,17 @@ def detect(cls, stream):
...
+        cls, dataset, freeze_panes=True, invalid_char_subst="-",
+        column_width: Optional[Union[str, int]] = "adaptive"
+    ):
         """Returns XLSX representation of Dataset.
 
         If dataset.title contains characters which are considered invalid for an XLSX file
         sheet name (http://www.excelcodex.com/2012/06/worksheets-naming-conventions/), they will
         be replaced with `invalid_char_subst`.
+
+        column_width: can have int, None, or "adaptive" as a value
```
hugovk, line 49:
```suggestion
        column_width: can be None, an integer, or "adaptive"
```

src/tablib/formats/_xlsx.py
```diff
@@ -51,6 +57,11 @@ def export_set(cls, dataset, freeze_panes=True, invalid_char_subst="-"):
         )
 
         cls.dset_sheet(dataset, ws, freeze_panes=freeze_panes)
+        if isinstance(column_width, str) and column_width != "adaptive":
+            raise ValueError(f"Unsupported value `{column_width}` passed to `column_width` "
+                             f"parameter. It supports 'adaptive' or integer values")
```
hugovk, line 62: No need for f-string here:
```suggestion
                             "parameter. It supports 'adaptive' or integer values")
```

hugovk (changes requested):
Some general comments. And please could we have three separate test cases that call `export_set` with different values for `column_width`: `None`, an integer and `"adaptive"`.

That way we see each branch of your code is running without error, and it will be reflected in the coverage.

Birdi7:
Thank you for such usefull comment! I've found a bug with new tests. Your comments are pushed, alongside with all three tests. What do you think on it?

tests/test_tablib.py
```diff
@@ -1100,6 +1102,67 @@ def test_xlsx_bad_dimensions(self):
...
+        with xls_source.open('rb') as fh:
+            data = tablib.Dataset().load(fh)
+        width_before = _get_width(data)
+        data.append([
+            'verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue',
+        ])
+        width_after = _get_width(data)
+
+        self.assertEqual(_some_integer, width_before)
+        self.assertEqual(_some_integer, width_after)
```
hugovk, line 1164: Thanks for the tests! 

How about this refactor that removes a lot of duplicate code into a helper function?

Also adds a fourth test case to check invalid input raises `ValueError`:

```suggestion
    def _helper_export_column_width(self, input_arg):
        """check that column width adapts to value length"""
        def _get_width(data, input_arg):
            xlsx_content = data.export('xlsx', column_width=input_arg)
            wb = load_workbook(filename=BytesIO(xlsx_content))
            ws = wb.active
            return ws.column_dimensions['A'].width

        xls_source = Path(__file__).parent / 'files' / 'xlsx_cell_values.xlsx'
        with xls_source.open('rb') as fh:
            data = tablib.Dataset().load(fh)
        width_before = _get_width(data, input_arg)
        data.append([
            'verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue-verylongvalue',
        ])
        width_after = _get_width(data, width_before)
        return width_before, width_after

    def test_xlsx_column_width_none(self):
        """check column width with None"""
        width_before, width_after = self._helper_export_column_width(None)
        self.assertEqual(width_before, 13)
        self.assertEqual(width_after, 13)

    def test_xlsx_column_width_adaptive(self):
        """check column width with 'adaptive'"""
        width_before, width_after = self._helper_export_column_width("adaptive")
        self.assertEqual(width_before, 11)
        self.assertEqual(width_after, 11)

    def test_xlsx_column_width_integer(self):
        """check column width with an integer"""
        width_before, width_after = self._helper_export_column_width(10)
        self.assertEqual(width_before, 10)
        self.assertEqual(width_after, 10)

    def test_xlsx_column_width_value_error(self):
        """check column width with invalid input"""
        with self.assertRaises(ValueError):
            self._helper_export_column_width("invalid input")
```

Birdi7:
Thank you for your comment! I believe this significantly  improves redability. Could please take a look if the PR is fine now?

src/tablib/formats/_xlsx.py
```diff
@@ -3,6 +3,7 @@
 
 import re
 from io import BytesIO
+from typing import Optional, Union
```
hugovk, line 6: We only support Python 3.7 and up.

Can we remove this import and instead use `|` instead of `Union` and use `| None` instead of `Optional`?

We might need `from __future__ import annotations` instead.

hugovk: This doesn't work with Python 3.7:
```console
$ cat 1.py
def thing(bleep: str | None):
    ...

$ python3.7 1.py
  File "1.py", line 1, in <module>
    def thing(bleep: str | None):
TypeError: unsupported operand type(s) for |: 'type' and 'NoneType'
$ python3.7 -m mypy 1.py
1.py:1: error: X | Y syntax for unions requires Python 3.10
Found 1 error in 1 file (checked 1 source file)
```

But this does:
```console
$ cat 1.py
from __future__ import annotations

def thing(bleep: str | None):
    ...

$ python3.7 1.py
$ python3.7 -m mypy 1.py
Success: no issues found in 1 source file
```

But you can keep the `typing` imports if you prefer :)

Birdi7: Thank you! I didn't get it for the first time. I've updated the typings with `from __future__ imports annotations` feature. Cound you check if this code is ready to be merged now?

docs/formats.rst
```diff
@@ -234,6 +234,19 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
...
+value you pass, the column width will be set accordingly. It can be either ``None``, an integer, or "adaptive".
+If "adaptive" is passed, the column width will be unique for every column and will be 
+calculated based on values' length. For example:: 
+
+    data = tablib.Dataset()
+    data.export('xlsx', column_width='adaptive')
+
+
+
+.. versionchanged:: 3.3.0
```
hugovk, line 247: 3.3.0 has been released, let's target 3.4.0:

```suggestion
.. versionchanged:: 3.4.0
```

src/tablib/formats/_xlsx.py
```diff
@@ -51,6 +57,11 @@ def export_set(cls, dataset, freeze_panes=True, invalid_char_subst="-"):
         )
 
         cls.dset_sheet(dataset, ws, freeze_panes=freeze_panes)
+        if isinstance(column_width, str) and column_width != "adaptive":
+            raise ValueError(f"Unsupported value `{column_width}` passed to `column_width` "
+                             "parameter. It supports 'adaptive' or integer values")
```
hugovk, line 62: Maybe something like this, to be closer to existing exceptions?

And let's start assigning exception messages to a variable first, for the reasons set out at https://github.com/henryiii/flake8-errmsg

```suggestion
            msg = (
                f"Invalid value for column_width: {column_width}. "
                f"Must be 'adaptive' or an integer."
            )
            raise ValueError(msg)
```

hugovk (changes requested):
Thank you! A couple of minor suggestions.

rakinson:
Hi there, I would find this enhancement useful. Do you mind to take this last step to add this feature? 
Thank you!

Birdi7:
Hi! I won't be able to finish this PR, so feel free to do that!

andrewgy8:
Hey! We think it would be useful over at Django-Import-Export!

I dont mind picking up this PR (rebasing, docs updates, and your comments @hugovk ) if we think it can be merged in. 

Given the work put in already by you all, it would be a shame to leave it abandoned. Thanks!

claudep:
Sure, feel free to take and polish this PR!

andrewgy8:
Okie dokie! I have:
- updated the code with comments from @hugovk ✅ 
- rebased on master ✅ 
- cleaned up the tests a bit ✅ 
- and cleaned up the column width code a bit ✅ 

Let me know if there is anything else I can do to get this across the line! Thanks for the help!

docs/formats.rst
```diff
@@ -250,6 +250,19 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending on the
+value passed, the column width will be set accordingly. It can be either ``None``, an integer, or default "adaptive".
```
claudep, line 254: We don't have a linter for the docs, but you may wrap this line in accordance of the rest of the file.

docs/formats.rst
```diff
@@ -250,6 +250,19 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
...
+If "adaptive" is passed, the column width will be unique and will be 
+calculated based on values' length. For example:: 
+
+    data = tablib.Dataset()
+    data.export('xlsx', column_width='adaptive')
+
+
+
```
claudep, line 262: I guess one line break would be enough here.

src/tablib/formats/_xlsx.py
```diff
@@ -166,3 +174,31 @@ def dset_sheet(cls, dataset, ws, freeze_panes=True, escape=False):
 
                 if escape and cell.data_type == 'f' and cell.value.startswith('='):
                     cell.value = cell.value.replace("=", "")
+
+    @classmethod
+    def _adapt_column_width(cls, worksheet, width):
+        if isinstance(width, str) and width != "adaptive":
+            msg = (
+                f"Invalid value for column_width: {width}. "
+                f"Must be 'adaptive' or an integer."
```
claudep, line 183: That second line doesn't need the `f` prefix.

src/tablib/formats/_xlsx.py
```diff
@@ -48,6 +48,12 @@ def export_set(cls, dataset, freeze_panes=True, invalid_char_subst="-", escape=F
         If ``escape`` is True, formulae will have the leading '=' character removed.
         This is a security measure to prevent formulae from executing by default
         in exported XLSX files.
+
+        If ``column_width`` is set to "adaptive", the column width will be set to the maximum
+        width of the content in each column. If it is set to an integer, the column width will be
+        set to that integer value. If it is set to None, the column width will be set as the
+        default openpyxl.Worksheet width value.
```
claudep, line 55: What do you think about accepting a list of column widths to be able to set widths per column? It may also be a follow-up improvement PR.

claudep (commented):
Thanks for refreshing this PR! You may also add an entry in HISTORY.md.

src/tablib/formats/_xlsx.py
andrewgy8: Yeah, I think that would be a valid iteration on this change 👍

andrewgy8:
Updated @claudep ! Thanks for the quick review!

claudep:
Thanks, let's wait now for @hugovk to make a final review.

docs/formats.rst
```diff
@@ -250,6 +250,18 @@ The ``import_set()`` method also supports a ``skip_lines`` parameter that you
 can set to a number of lines that should be skipped before starting to read
 data.
 
+The ``export_set()`` method supports a ``column_width`` parameter. Depending 
+on the value passed, the column width will be set accordingly. It can be 
+either ``None``, an integer, or default "adaptive". If "adaptive" is passed, 
+the column width will be unique and will be calculated based on values' length. 
+For example::
```
hugovk, line 257: Let's remove those trailing spaces:
```suggestion
The ``export_set()`` method supports a ``column_width`` parameter. Depending
on the value passed, the column width will be set accordingly. It can be
either ``None``, an integer, or default "adaptive". If "adaptive" is passed,
the column width will be unique and will be calculated based on values' length.
For example::
```

src/tablib/formats/_xlsx.py
```diff
@@ -166,3 +174,31 @@ def dset_sheet(cls, dataset, ws, freeze_panes=True, escape=False):
...
+        column_widths = []
+        if width == "adaptive":
+            for row in worksheet.values:
+                for i, cell in enumerate(row):
+                    cell = str(cell)
+                    if len(column_widths) > i:
+                        if len(cell) > column_widths[i]:
+                            column_widths[i] = len(cell)
+                    else:
+                        column_widths.append(len(cell))
```
hugovk, line 199: We're not using `cell` for anything else than `len(cell)`,  so perhaps something like this?
```suggestion
                    cell_width = len(str(cell))
                    if len(column_widths) > i:
                        if cell_width > column_widths[i]:
                            column_widths[i] = cell_width
                    else:
                        column_widths.append(cell_width)
```

andrewgy8: Good catch!

hugovk (approved):
Thanks!

andrewgy8:
Thanks all! Looking forward to adding this to Django Import Export!

hugovk:
I'll prepare a new release!

Issue #517, opened 2022-02-21
Support Mac OS LF char for csv


matthewhegarty:
Hi
I'm testing some changes on `django-import-export` and I notice an issue with LF endings for csv data, which was the format on [older mac os distributions](https://community.perforce.com/s/article/3096#:~:text=Platform%20Conventions&text=On%20Windows%2C%20line%2Dendings%20are,(%5Cr%20or%20CR).).

```python
>>> import tablib
>>> sb = 'id,name,author_email\r1,Some book,test@example.com\r'
>>> tablib.import_set(sb, format='csv')
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "/home/matthew/.virtualenvs/django-import-export/lib/python3.9/site-packages/tablib/core.py", line 908, in import_set
    return Dataset().load(normalize_input(stream), format, **kwargs)
  File "/home/matthew/.virtualenvs/django-import-export/lib/python3.9/site-packages/tablib/core.py", line 414, in load
    fmt.import_set(self, stream, **kwargs)
  File "/home/matthew/.virtualenvs/django-import-export/lib/python3.9/site-packages/tablib/formats/_csv.py", line 44, in import_set
    for i, row in enumerate(rows):
_csv.Error: new-line character seen in unquoted field - do you need to open the file in universal-newline mode?
```
This can be fixed by adding the ['newline'](https://docs.python.org/3/library/io.html#io.StringIO) kwarg to `normalize_input()`:

```python
def normalize_input(stream):
    """
    Accept either a str/bytes stream or a file-like object and always return a
    file-like object.
    """
    if isinstance(stream, str):
        return StringIO(stream, newline='')
    elif isinstance(stream, bytes):
        return BytesIO(stream)
    return stream
```
- tablib: 3.2.0
- python 3.9.10

matthewhegarty:
@claudep Is it possible we could get a new release with this fix in?  We would need it for our forthcoming v3 release of `django-import-export`.  I also notice that there was a [build failure](https://github.com/jazzband/tablib/actions/runs/1877744964).  Could you re-run the job to test whether it is something we need to fix?

hugovk:
Yes, it's about time for a release. We could wait a day or two to see if we can get https://github.com/jazzband/tablib/pull/516 in?

What's your schedule for `django-import-export`?

I don't want to delay if you're ready to go now, we can always release this now and make a new release when that's ready, it's pretty easy to release with the automation.

---

[...]

PR #518, opened 2022-02-21
Add newline to StringIO

matthewhegarty:
This is a fix for #517.
Adding the [newline](https://docs.python.org/3/glossary.html#term-universal-newlines) kwarg to `StringIO` handles interpreting the LF char.

https://docs.python.org/3/library/functions.html#open-newline-parameter

### DIFF
--- a/HISTORY.md
+++ b/HISTORY.md
@@ -1,5 +1,11 @@
 # History
 
+## 3.2.1 (unreleased)
+
+### Bugfixes
+
+- Support solo CR in text input imports (#518).
+
 ## 3.2.0 (2022-01-27)
 
 ### Changes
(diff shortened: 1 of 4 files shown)


claudep:
Thanks! May I ask you to also add a line in `HISTORY.md`?

matthewhegarty:
Have updated HISTORY as requested.

HISTORY.md
```diff
@@ -1,5 +1,11 @@
 # History
 
+## 3.2.1 (unreleased)
+
+### Bugfixes
+
+- Add newline kwarg to `StringIO` (#518).
```
claudep, line 7: Thanks, but I would rather describe the functional benefit, something like `Support solo CR in text input imports.`

matthewhegarty: Done

claudep (approved):

PR #519, opened 2022-04-01
CI: Add workflow_dispatch to add a UI button for triggering builds

hugovk:
This adds a button for manually starting builds, like Travis CI has.

Also update some actions to latest v3, and update pre-commit.


claudep (approved):
:+1:

PR #520, opened 2022-04-03
Upgrade unit tests

hugovk:
I used https://github.com/isidentical/teyit to upgrade the unit tests to use methods which give more useful error messages upon failure.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -9,7 +9,11 @@ repos:
     rev: 5.10.1
     hooks:
       - id: isort
-        additional_dependencies: [toml]
+
+  - repo: https://github.com/isidentical/teyit
+    rev: 0.4.2
+    hooks:
+      -   id: teyit
 
   - repo: https://github.com/pre-commit/pygrep-hooks
     rev: v1.9.0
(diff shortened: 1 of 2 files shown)


claudep (approved):
Looks good!

hugovk:
The "pre-commit.ci - pr" check is failing because the pre-commit is checking against master instead of a tag here: https://github.com/jazzband/tablib/pull/520/files#diff-63a9c44a44acf85fea213a857769990937107cf072831e1a26808cfde9d096b9R14

We need to use `master` because the latest tag doesn't have this fix: https://github.com/isidentical/teyit/pull/14

Let's just remove the pre-commit check for now, and can add it back when there's a release.

isidentical:
I've released 0.4.2 with the changes in the pre-commit hook: https://github.com/isidentical/teyit/releases/tag/0.4.2

hugovk:
Thanks for the new release!

"pre-commit.ci - pr" now errors because it seems it's running on Python 3.8. I've opened an issue to ask how we can set a Python version at pre-commit.ci: https://github.com/pre-commit-ci/issues/issues/117

PR #525, opened 2022-09-28
Add support for Python 3.11

hugovk:
[Python 3.11.0 final](https://peps.python.org/pep-0664/) is set for release on 2022-10-24, we can already test and set the Trove classifier.

Also bump GitHub Actions versions.


claudep (approved):
Good!

PR #528, opened 2022-10-30
Fixes #527 - Export integers/floats as numbers in ODS format

### LINKED ISSUE #527: ODS exports numbers as text?
I've been trying to export data with numbers in them, and found I could not do calculations with them in the resulting ODS spreadsheet (the values are left-aligned instead of right aligned, and have a ' in front of them).

Here's my testcase:

```
data = tablib.Dataset()
data.append((1, 2.1))
f = open('tmp.ods', 'w')
f.write(data.ods)
```

Opening the resulting ODS in libreoffice shows:

image

Note the leading `'` in the field value. I suspect that this leading `'` is not written by tablib itself, but tablib writes out table cells as text. Here's a snippet from `content.xml` in the generated ODS:

    <table:table-cell><text:p>1</text:p></table:table-cell><table:table-cell><text:p>2.1</text:p></table:table-cell>

The above screenshot is with the libreoffice locale set to "English (USA)". I originallly had it set to "Dutch (Netherlands)", which uses `,` as the decimal separator, so I suspected it was a decimal separator problem, but switching to a local with `.` as the decimal separator did not help.

Exporting in the xlsx format does give a proper numerical value that can be used in calculations (with both locales), so I'll be using that as a workaround for now.

claudep:


src/tablib/formats/_ods.py
```diff
@@ -53,45 +53,18 @@ def dset_sheet(cls, dataset, ws):
...
-                    odf_row.setAttribute('stylename', bold)
-                    ws.addElement(odf_row)
-                    cell = table.TableCell()
-                    p = text.P()
-                    p.addElement(text.Span(text=col, stylename=bold))
-                    cell.addElement(p)
-                    odf_row.addElement(cell)
-
-                # wrap the rest
+                if isinstance(col, (int, float)):
```
matthijskooijman, line 62: In my original usecase, I was using `decimal.Decimal` values, which are not handled by your code now. Changing this line as shown below should catch all numeric values (as [suggested here](https://stackoverflow.com/a/4187266/740048)) and makes your PR work perfectly for my case with `Decimal` values (of course the import should go at the top).

```suggestion
                import numbers
                if isinstance(col, numbers.Number):
```

matthijskooijman (changes requested):
Thanks for this fix, it seems to work as expected. I left one comment inline to make it even more generic.

matthijskooijman (approved):

src/tablib/formats/_ods.py
claudep: Excellent suggestion, thanks!

tests/test_tablib.py
```diff
@@ -990,6 +991,15 @@ def test_tsv_export(self):
         self.assertEqual(tsv, self.founders.tsv)
 
 
+class ODSTests(BaseTestCase):
+    def test_ods_export_datatypes(self):
+        date_time = datetime.datetime(2019, 10, 4, 12, 30, 8)
+        data.append(('string', '004', 42, 21.55, Decimal('34.5'), date_time))
+        data.headers = ('string', 'start0', 'integer', 'float', 'decimal', 'date/time')
+        # ODS is currently write-only, just test that output doesn't crash.
+        data.ods
```
hugovk, line 1000: Can we assert something on this?

Even if it's something like `assert data.ods is not None` and/or `assert len(data.ods)`?

```suggestion
        assert data.ods is not None
        assert len(data.ods)
```

src/tablib/formats/_ods.py
```diff
@@ -53,45 +54,18 @@ def dset_sheet(cls, dataset, ws):
...
-                    ws.addElement(odf_row)
-                    cell = table.TableCell()
-                    p = text.P()
-                    p.addElement(text.Span(text=col, stylename=bold))
-                    cell.addElement(p)
-                    odf_row.addElement(cell)
-
-                # wrap the rest
+                if isinstance(col, numbers.Number):
+                    tc = table.TableCell(valuetype="float", value=col)
```
hugovk, line 64: Can we keep this named more descriptively as `cell` like elsewhere?

```suggestion
                    cell = table.TableCell(valuetype="float", value=col)
```

src/tablib/formats/_ods.py
```diff
@@ -53,45 +54,18 @@ def dset_sheet(cls, dataset, ws):
...
-                            cell.addElement(text.P(text=col))
-                            odf_row.addElement(cell)
-                    except TypeError:
-                        ws.addElement(odf_row)
-                        cell = table.TableCell()
-                        cell.addElement(text.P(text=col))
-                        odf_row.addElement(cell)
+                    tc = table.TableCell(valuetype="string")
+                    tc.addElement(text.P(text=str(col), stylename=style))
+                odf_row.addElement(tc)
```
hugovk, line 68:
```suggestion
                    cell = table.TableCell(valuetype="string")
                    cell.addElement(text.P(text=str(col), stylename=style))
                odf_row.addElement(cell)
```

hugovk:
GitHub tip: put "Fixes https://github.com/jazzband/tablib/issues/527" in the PR message and it will autoclose the issue when merged.

It doesn't trigger from the title.

hugovk (approved):
Thanks!

PR #535, opened 2023-01-21
fix bug when checking empty file. add more coverage

llazzaro:
While adding some coverage discovered the using `len(pickle)` will raise and error since for empty files `pickle` is `None`.

Change the if condition to just `not pickle` since it will also work for checking if the list is empty.

Added two test cases to cover more branches.

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -334,7 +334,8 @@ class Dataset:
 
         """
         error_details = "Please check format documentation https://tablib.readthedocs.io/en/stable/formats.html#yaml"
-        if not len(pickle):
+
+        if not pickle:
             return
         
         if not isinstance(pickle, list):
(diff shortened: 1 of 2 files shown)


claudep (approved):
Seems good, thanks!

claudep:
Maybe just adapt the commit message to specify this concerns the yaml format.

hugovk (changes requested):
Please could you check the permissions failure on Windows?

claudep:
Ah yes, NamedTemporaryFile constructor is already returning a file-like object, so no need to reopen later (and Windows cannot reopen when it's already open).

llazzaro:
* git commit message fixed
* fix windows broken test

hugovk (approved):
Thanks!

PR #537, opened 2023-02-08
Add flake8 hook to pre-commit

peymanslh:
Add flake8 to pre-commit and remove commented and old configs from the tox file.


hugovk:
We need to update `docs/conf.py`, something like:

```diff
 # The encoding of source files.
-#source_encoding = 'utf-8-sig'
+# source_encoding = 'utf-8-sig'
```

peymanslh:
> We need to update `docs/conf.py`, something like:
> 
> ```diff
>  # The encoding of source files.
> -#source_encoding = 'utf-8-sig'
> +# source_encoding = 'utf-8-sig'
> ```

Yes, I will fix it.
But I excluded `docs/` in the flake8 config!(I don't know why it's not working :D)

hugovk:
I think it's because pre-commit ignores the `exclude` list and scans all files in Git.
 
Let's include the `docs` directory, `conf.py` is a Python file after all so let's scan it too.

hugovk (approved):
Thanks!

Issue #538, opened 2023-02-13
Add <tbody> tags to html output


krowbar:
A table exported to html with a header row no longer renders properly (on recent chrome, edge and firefox) if passed through a minifier that drops optional end tags.
The table renders styled as all header rows.

Wrapping the rows with tbody.open/tbody.close fixes the rendering problem.

` page.tbody.open()`
` for row in dataset:`
`    new_row = [item if item is not None else "" for item in row]`
`    html_row = markup.oneliner.td(new_row)`
`    page.tr(html_row)`
`page.tbody.close()`

In the case I noticed, the table was 12K minified, but 19K minified if the end tags were included -- about a hundred rows and 9 columns.

krowbar:
[tablib_html.patch](https://github.com/jazzband/tablib/files/10746830/tablib_html.patch)

Let me know if you want something different.

claudep:
Ideally if you could make it a pull request, would be nice! Tell us if you need any help.

krowbar:
Pull request submitted.

krowbar:
Updated the html test generation in my version.

PR #539, opened 2023-02-16
Add <tbody> tags to html output

krowbar:
Resolves jazzband/tablib#538

Wrap data rows in `<tbody>` tags to fix minifying (with a header row) when dropping optional end tags.
Otherwise the entire table renders with `<thead>` styling.


claudep:
Looks like tests need to be adapted.

claudep (approved):
Thanks!

Issue #257, opened 2016-11-25
Allow escaping of excel formulas


czpython:
Hello,
Packages that leverage this app like django-tablib allows us to export user submitted data.
This exposes a potential security issue where users can add excel formulas to certain exported fields.

Excel allows us to use ``'`` at the beginning of any cell value to have it treated as text only and thus not execute any formula.

I did some digging and looks like this feature is not supported.
I'm interested in contributing this but I'm not 100% sure exactly where, is this something that should go on the packaged dependencies?

iurisilvio:
I think it can be done in xlsx formatter. https://github.com/kennethreitz/tablib/blob/master/tablib/formats/_xlsx.py#L123

You can add a `safe` argument to `import_book`/`import_set` and change the import behaviour when `safe=True`.

Pull requests are welcome. :+1:

matthewhegarty:
I raised this on the [openpyxl forum](https://groups.google.com/g/openpyxl-users/c/dQdn9NnIcg0/m/05WCSh4pMQAJ?utm_medium=email&utm_source=footer) to ask about best practice for escaping formula.

The response was to rewrite cell contents to remove the '=' char for function cell types.

What is the opinion here about implementing this in tablib? (we would potentially need this downstream in django-import-export).  I would be happy to work on a PR if you think this is suitable.

claudep:
I'll gladly review a PR for this.

matthewhegarty:
@claudep Have submitted PR #540 - would be grateful if you could take a look

PR #540, opened 2023-03-01
Escape formulae on export

matthewhegarty:
There is a potential security flaw where XLSX files can be exported with active formulae (see #257).

This PR adds an optional 'escape' kwarg to the export method which will strip leading '=' characters and disable formulae.

Tests included.


tests/test_tablib.py
```diff
@@ -1117,6 +1118,68 @@ def test_xlsx_cell_values(self):
...
+        """
+        data.append(('=SUM(1+1)',))
+        _xlsx = data.export('xlsx')
+
+        # read back using openpyxl because tablib reads formulae as values
+        wb = load_workbook(filename=BytesIO(_xlsx))
+        ws = wb.active
+        self.assertEqual('=SUM(1+1)', ws['A1'].value)
+
+        _xlsx = data.export('xlsx', escape=True)
```
hugovk, line 1133: Why the leading underscores on these?

tests/test_tablib.py
```diff
@@ -1117,6 +1118,68 @@ def test_xlsx_cell_values(self):
...
+        wb = load_workbook(filename=BytesIO(_xlsx))
+        ws = wb.active
+        self.assertEqual('SUM(1+1)', ws['A1'].value)
```
hugovk, line 1136: Let's use `wb` directly and avoid a temp variable only used once:

```suggestion
        self.assertEqual('SUM(1+1)', wb.active['A1'].value)
```

tests/test_tablib.py
matthewhegarty: Thanks for reviewing Hugo.  I used the leading underscore to follow the existing convention used elsewhere in the test file.

tests/test_tablib.py
matthewhegarty: Done

tests/test_tablib.py
hugovk: Ah right, let's stay consistent :)

claudep:
Thanks for the patch! Could you please also complete the docs in `formats.rst`?

matthewhegarty:
> Thanks for the patch! Could you please also complete the docs in `formats.rst`?

Done.  To avoid duplication I added links to the source docstrings.

Also updated requirements.txt to the latest version of 'Sphinx'

claudep (approved):
That looks very good, thanks!

claudep:
@hugovk, I let you do the merge if you find it OK.

hugovk:
Thanks!

matthewhegarty:
Is there a timeline for releasing this?   Sorry to ask but we have a downstream security issue relating to this.  Any updates greatly appreciated.

claudep:
I guess someone has to complete the HISTORY file, but then I think we could release soonish...

hugovk:
@matthewhegarty If you can make a PR to update `HISTORY.md`, that would help us release much quicker :)

matthewhegarty:
#544 :+1:

PR #541, opened 2023-03-04
xlsx export: remove redundant code

matthewhegarty:
This PR removes what I believe is a redundant check from XLSX export `dset_sheet()`.  It provides a minor simplification of the code.

If we check the source for `openpyxl.cell.cell.Cell`, the cell value setter calls `_bind_value()`.  This method can raise a `ValueError` but not a `TypeError`.  Therefore I have removed the checks for `TypeError`, and I assume that if this does occur somehow, then we would want this raised and not caught because this might be masking an error the user might be interested in.

I also removed catching `TypeError` when `str(col)` is called for the same reason.

I have added tests to increase the coverage.


tests/test_tablib.py
```diff
@@ -1180,6 +1182,18 @@ def test_xlsx_bad_dimensions(self):
...
+        """Test that the process handles errors which might be raised
+        when calling cell setter."""
+        setter_mock = Mock(wraps=Cell.value.fset, side_effect=[ValueError, '1'])
+        mock_property = Cell.value.setter(setter_mock)
+        with patch.object(Cell, 'value', mock_property):
+            data.append((1,))
+            _xlsx = data.export('xlsx')
+            wb = load_workbook(filename=BytesIO(_xlsx))
+            # note the Cell is mocked so we don't get the 'real' value
+            self.assertEqual(None, wb.active['A1'].value)
```
claudep, line 1195: Do you think we could try with a value really producing a `ValueError` instead of this mocking/patching dance?

matthewhegarty: ok - have used an 'array' value which openpyxl cannot handle.  This is probably clearer but means we are also testing openpyxl code, instead of mocking as an external dependency.  So if they ever change their implementation to handle arrays, this test will break.

claudep: I see, but anyway I tend to favor this simpler version. I think the former mocking solution was also sensible to implementation changes. @hugovk might arbitrate that choice :smile:

hugovk: Also prefer this simpler version, we'll just have to update the test if they change the implementation.

matthewhegarty: Thanks for reviewing guys - happy to go with the simpler approach.  Over to you if you are happy to merge.

claudep (approved):

PR #542, opened 2023-03-04
Move more setup stuff to pyproject.toml

claudep:


tox.ini
```diff
@@ -24,9 +24,10 @@ commands =
 [testenv:lint]
 deps =
     pre-commit
+    build
     twine
 commands =
     pre-commit run --all-files
-    python setup.py sdist
+    python -m build --sdist
```
hugovk, line 31: Let's generate and check the wheel too:

```suggestion
    python -m build
```

hugovk (commented):
Nice! I think we can now delete `setup.py` completely?

We should also change this in the three workflows:

```diff
-cache-dependency-path: "setup.py"
+cache-dependency-path: "pyproject.toml"
```

claudep:
> Nice! I think we can now delete `setup.py` completely?

I read that it might be useful for some older environments, but if you think it's not needed, we can remove it too (and see if we get complaints :wink:).

hugovk:
Let's try without the stub `setup.py`:

https://setuptools.pypa.io/en/latest/userguide/quickstart.html#basic-use says (emphasis added):

> We also recommend users to expose as much as possible configuration in a more declarative way via the [pyproject.toml](https://setuptools.pypa.io/en/latest/userguide/pyproject_config.html) or [setup.cfg](https://setuptools.pypa.io/en/latest/userguide/declarative_config.html), and keep the setup.py minimal with only the dynamic parts (_or even omit it completely if applicable_).

And https://setuptools.pypa.io/en/latest/userguide/quickstart.html#development-mode says:

> Prior to [pip v21.1](https://pip.pypa.io/en/latest/news/#v21-1), a setup.py script was required to be compatible with development mode. With late versions of pip, projects without setup.py may be installed in this mode.
> 
> If you have a version of pip older than v21.1 or is using a different packaging-related tool that does not support [PEP 660](https://peps.python.org/pep-0660/), you might need to keep a setup.py file in file in your repository if you want to use editable installs.

If anyone needs editable installs, I think it's fair enough to require modern pip (= less than two years old).

We can always add the stub back if it's really needed.

claudep:
OK, fine. setup.py removed!

hugovk (approved):
🚀

PR #547, opened 2023-04-07
Allow importing 'ragged' .xlsx files through dataset

### LINKED ISSUE #465: tablib no longer able to import Google Docs exported xlsx

clemencealainpadam:
Fixes #465.

This is an almost blind copy paste of the work done by @claudep in #466 to allow importing ragged file through dataset also.
(https://github.com/jazzband/tablib/issues/465#issuecomment-1172260736)

### DIFF
--- a/src/tablib/formats/_xlsx.py
+++ b/src/tablib/formats/_xlsx.py
@@ -85,13 +85,8 @@ class XLSXFormat:
         return stream.getvalue()
 
     @classmethod
-    def import_set(cls, dset, in_stream, headers=True, read_only=True, skip_lines=0):
-        """Returns databook from XLS stream."""
-
-        dset.wipe()
-
-        xls_book = load_workbook(in_stream, read_only=read_only, data_only=True)
-        sheet = xls_book.active
+    def import_sheet(cls, dset, sheet, headers=True, skip_lines=0):
+        """Populates dataset with sheet."""
 
         dset.title = sheet.title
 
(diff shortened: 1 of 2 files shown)


claudep:
Good catch! I wonder if it would be worth factorizing that sheet import loop between both methods so such mistakes are less probable in the future? What do you think?

clemencealainpadam:
Agreed ! I propose a new function `import_sheet` [1e37bce](https://github.com/jazzband/tablib/pull/547/commits/1e37bce145d8cb4e28e23a1730f6759819ad1092)

claudep (approved):
Thanks a lot!

PR #550, opened 2023-06-08
Test Python 3.12 beta

hugovk:
The [Python 3.12 beta](https://discuss.python.org/t/python-3-12-beta-1-and-feature-freeze-is-here/26982?u=hugovk) is now out, let's start testing it: https://dev.to/hugovk/help-test-python-312-beta-1508/

Because pandas doesn't yet support 3.12, we need to skip the pandas-related tests for now. We can do that with `pip install --only-binary :all: pandas`, which only installs pandas if there's a prebuilt binary wheel available (as it is for 3.7-3.11), and skip when it's not available.

I didn't add the 3.12 Trove classifier yet. We could wait until full release in October, but some other projects have added it already:

* 7 in the top 360: https://pyreadiness.org/3.12/
* 1,083 in total: https://pypi.org/search/?c=Programming+Language+%3A%3A+Python+%3A%3A+3.12

Shall we wait or add it now?

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
+        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11", "3.12"]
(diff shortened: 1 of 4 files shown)


claudep (approved):
No strong opinion regarding the trove classifier.

hugovk:
I've added it then, we don't need to remember to do so later :)

PR #551, opened 2023-06-08
Drop support for EOL Python 3.7

hugovk:
Python 3.7 is EOL this month:

image

* https://peps.python.org/pep-0537/

Let's drop it. I don't mind if we wait until 2023-06-27, but we're probably close enough?

---

Does dropping a Python warrant a major or minor bump?

I used to do major, but `python_requires = >=3.8` ensures the new release only installs for the right group, and I'm now leaning towards a minor, but either is fine with me.

(If we had deprecations to drop, I'd do those at the same time and do major, but we don't.)

A [recent poll](https://hachyderm.io/@nedbat/110412912217669532) on the topic favours minor:

image

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -11,7 +11,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.7", "3.8", "3.9", "3.10", "3.11"]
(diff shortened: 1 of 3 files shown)


claudep (approved):
Sorry, once again no strong opinion about minor vs major. Minor is fine for me.

hugovk:
Let's do minor then :)

PR #554, opened 2023-07-01
Replace MarkupPy by ElementTree for html conversion

claudep:


mgorny:
Thanks. That's definitely better idea than mine (of skipping HTML tests). I also like that you're using DOM to create elements, that's probably the most efficient way of doing it.

hugovk (approved):

hugovk:
Thanks!

claudep:
Thanks for the merge.

While I was on this format, I explored a bit about adding html import support, also without external dependency: #555

matthewhegarty:
I noticed that this change [breaks the CI build](https://github.com/django-import-export/django-import-export/issues/1602) in django-import-export.  This is for cases when we are escaping harmful characters such as script tags.

If there is already an escaped character in content, then the use of ElementTree will double escape, which will not format correctly. This test will reproduce the issue:

```python
    def test_html_export_with_special_chars(self):
        self.founders = tablib.Dataset(headers=self.headers, title='Founders')
        self.founders.append(('J &amp; J', 'A', 90))
        self.assertIn("J &amp; J", self.founders.html)
```

This test fails because the '&amp;' string is exported as '&amp;amp;'

claudep:
@matthewhegarty Please create a new issue for this problem where we can discuss that double-escaping issue.

PR #557, opened 2023-07-19
Get row method

### LINKED ISSUE #24: get_row, get_col methods

yoonthegoon:
Fixes #24 get_row, get_col methods

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -32,3 +32,4 @@ Here is a list of past and present much-appreciated contributors:
     Tommy Anthony
     Tsuyoshi Hombashi
     Tushar Makkar
+    Yunis Yilmaz
--- a/docs/tutorial.rst
+++ b/docs/tutorial.rst
@@ -167,7 +167,13 @@ You can slice and dice your data, just like a standard Python list. ::
 
     >>> data[0]
     ('Kenneth', 'Reitz', 22)
+    >>> data[0:2]
+    [('Kenneth', 'Reitz', 22), ('Bessie', 'Monke', 20)]
 
+You can also access a row using its index without slicing. ::
+
+    >>> data.get(0)
+    ('Kenneth', 'Reitz', 22)
 
 If we had a set of data consisting of thousands of rows,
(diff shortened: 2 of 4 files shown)


yoonthegoon:
Would one of you please take a look when you get the chance?
I'm looking to try to clean out some old issues on here when I get time. @hugovk @claudep

hugovk:
Thanks for the PR.

https://github.com/jazzband/tablib/pull/28 added `get_col`, should this be named `get_row` instead of `get` (as suggested in https://github.com/jazzband/tablib/issues/24)?

yoonthegoon:
> Thanks for the PR.
> 
> 
> 
> https://github.com/jazzband/tablib/pull/28 added `get_col`, should this be named `get_row` instead of `get` (as suggested in https://github.com/jazzband/tablib/issues/24)?

i was originally gonna, but all the row methods in the class don't have `row` in their names while all the column methods do have `col`. kenneth also commented on the original issue that the methods he wanted were `get` to return a row by index and `get_col` for column by index.

hugovk (approved):
Ah yes, good point!

Thank you!

claudep:
Did you check if the docs should be updated somewhere?

yoonthegoon:
I built the docs locally and it looks like it added it by me, so the existing rst files appear to pick it up.
`docs/_build/*` is in the `.gitignore` so I don't think there's any need for me to commit a build.
Built docs are [here](https://tablib--557.org.readthedocs.build/en/557/api.html#tablib.Dataset.get) and look good to me.

hugovk:
We could add an example to the quickstart guide:

https://tablib.readthedocs.io/en/stable/tutorial.html#selecting-rows-columns

claudep:
The "add type validation" commit should not be part of this PR.

yoonthegoon:
> The "add type validation" commit should not be part of this PR.

Yeah I just noticed it changed more than I intended. 
Lemme fix that real quick

claudep:
:medal_sports:

PR #560, opened 2023-09-05
feat: `transpose` support dataset with no headers

sgqy:
`data.transpose()` raises `TypeError: 'NoneType' object is not subscriptable`, when `headers` is not set.

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -688,14 +688,10 @@ class Dataset:
 
         return _dset
 
-    def transpose(self):
+    def _transpose_with_headers(self):
         """Transpose a :class:`Dataset`, turning rows into columns and vice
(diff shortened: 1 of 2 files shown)


src/tablib/core.py
```diff
@@ -716,6 +712,35 @@ def transpose(self):
             _dset.append(row=row_data)
         return _dset
 
+    def _transpose_without_headers(self):
+        """Transpose a :class:`Dataset`, turning rows into columns and vice
+        versa, returning a new ``Dataset`` instance. This instance should not
+        has headers, or the dimension would be invalid."""
```
claudep, line 718: has -> have

claudep (commented):
Thanks! Almost ready, just a small typo.

claudep (approved):

Issue #558, opened 2023-07-23
Addition of ElementTree changes HTML output


matthewhegarty:
I noticed that [this change](https://github.com/jazzband/tablib/pull/554) [breaks the CI build](https://github.com/django-import-export/django-import-export/issues/1602) in django-import-export.  This is for cases when we are escaping harmful characters such as script tags.

If there is already an escaped character in content, then the use of ElementTree will double escape, which will not format correctly. This test (added to `HTMLTests`) will reproduce the issue:

```python
    def test_html_export_with_special_chars(self):
        self.founders = tablib.Dataset(headers=self.headers, title='Founders')
        self.founders.append(('J &amp; J', 'A', 90))
        self.assertIn("J &amp; J", self.founders.html)
```

(this test passes in v3.3.0)

This test fails because the '&amp;' string is exported as '&amp;amp;'  So if someone has 'J &amp; J' in their content (i.e. rendered in escape `& amp;`), they will now get "double escaping" 'J & amp;amp; J'.

claudep:
I understand this is a breaking change (if not reverted, it should be documented). However I think that security-wise, it's a much better default behavior than the previous one. Maybe we should add an option to the format so that the client code can specify that it already cared for escaping the content (and thus avoid the double-escaping)?

matthewhegarty:
As long as the escaping is done we can handle it ok (so no need for an additional option), however it would be ideal if we could opt into the new ElementTree format when we are ready.  Perhaps to do that you could add a deprecation path using a settings flag so that clients can opt into ElementTree?  I think without this you could well get complaints that exports have changed when the next release goes out.  Then it could be documented and the change explained (it took me a little while to figure out why our CI build was broken).

matthewhegarty:
Hi Claude - are you ok with me submitting a PR to try to make this new dependency on ElementTree optional?  Then we can plan the deprecation in django-import-export.  Please let me know and I'll pull it together.

claudep:
Submitting PRs doesn't require any permission :smile:

matthewhegarty:
Understood... Just wanted to double-check it was something you would consider before I started.

matthewhegarty:
I looked into whether we could control escaping of text elements using a parameter or similar, but this is not possible ([certain characters always get escaped](https://github.com/python/cpython/blob/baa6dc8e388e71b2a00347143ecefb2ad3a8e53b/Lib/xml/etree/ElementTree.py#L1013)).

[...]

PR #562, opened 2023-09-11
refactored html export to allow for optional escaping

matthewhegarty:
Refactored HTML export to write a table without the need for any external dependency.  Dataset values can be optionally encoded.  This is to address issue #558.


matthewhegarty:
Closing as per [comment](https://github.com/jazzband/tablib/issues/558#issuecomment-1715157090)

PR #566, opened 2023-10-26
fix(ods): None values are not displayed as string

hoccau:
Actually, `None` values are converted in string in an ODS sheet (so 'None' string is displayed in cells). This fill a cell empty value in case of None.
Sorry for not providing test : I see that ODS is tested only globally. It could be improved later :)
Thanks!


claudep:
I would suggest waiting for PR #568 to be merged, then you can rebase and take advantage of the new import functionality to complement the tests.

hoccau:
> I would suggest waiting for PR #568 to be merged, then you can rebase and take advantage of the new import functionality to complement the tests.

ok, fine.

claudep (approved):
Thanks! That looks good.

PR #568, opened 2023-10-27
 Fixes #567 - Implement ods import

claudep:

PR #569, opened 2023-10-29
Refactor `import datetime` as `import datetime as dt`

hugovk:
Follow on from https://github.com/jazzband/tablib/pull/568#discussion_r1375298382.

Refactor `import datetime` as `import datetime as dt` to avoid ambiguity between the `datetime` module and class:

* https://adamj.eu/tech/2019/09/12/how-i-import-pythons-datetime-module/

### DIFF
--- a/tests/test_tablib.py
+++ b/tests/test_tablib.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 """Tests for Tablib."""
 
-import datetime
+import datetime as dt
 import doctest
 import json
 import pickle
@@ -304,8 +304,8 @@ class TablibTestCase(BaseTestCase):
         """Passes in a single datetime and a single date and exports."""
 
         new_row = (
-            datetime.datetime.now(),
-            datetime.datetime.today(),
+            dt.datetime.now(),
+            dt.datetime.today(),
         )
 
         data.append(new_row)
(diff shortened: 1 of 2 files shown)


claudep:
I'm a bit hesitant to touch dbfpy too much, as it is vendored code. I we were to update the vendored code to a more recent version, we would face a big diff. What do you think?

hugovk:
Good point, I'll revert.

Shall we also rename the `package` directory to make it more obvious this is vendored?

For example, pip has `_vendor`:

* https://github.com/pypa/pip/tree/main/src/pip/_vendor

claudep (approved):

claudep:
> Shall we also rename the package directory to make it more obvious this is vendored?

Makes sense, as long as this doesn't introduce compatibility issues (but it should not at first sight).

hugovk:
> > Shall we also rename the package directory to make it more obvious this is vendored?
> 
> Makes sense, as long as this doesn't introduce compatibility issues (but it should not at first sight).

Hmm, I guess in theory someone could be doing `from tablib.packages import dbfpy`...

claudep:
That could happen, but frankly I don't see the use case. I let you judge if moving the package is worth it, considering that minor breakage.

hugovk:
Normally this sort of thing should be preceded by a deprecation warning, but I think we're fine here.

And the next release is set to be a major release anyway for https://github.com/jazzband/tablib/issues/558, so let's go for it.

PR #570, opened 2023-10-30
Move `src/tablib/packages/dbfpy` to `src/tablib/_vendor/dbfpy`

hugovk:
Follow on from https://github.com/jazzband/tablib/pull/569.

`dbfpy` is a vendored package. Move it to a `_vendor` directory to make this clear.

This will be released in tablib 4.0.0.


claudep (approved):
Thanks!

PR #573, opened 2023-12-19
Fixes #572 - Allow appending rows after a dynamic column was inserted

### LINKED ISSUE #572: appending rows when there are dynamic columns
At the moment we cannot add rows (using append, rpush, lpush) when there is already a dynamic column: a function object can be added at the appropriate place in the row, but unlike a normal dynamic column it will not be executed.

claudep:


src/tablib/core.py
```diff
@@ -547,6 +553,7 @@ def insert_col(self, index, col=None, header=None):
 
         # Callable Columns...
         if hasattr(col, '__call__'):
+            self._dyn_columns[self.width] = col
```
hugovk, line 556:
```suggestion
            self._dynamic_columns[self.width] = col
```

hugovk (approved):
Looks good, couple of small suggestions.

claudep:
Thanks for the review, but I'd like to work on it some more, specifically to allow the use case of people inserting complete new rows with dynamic values already calculated, and add some docs about it.

claudep:
A second quick review would be great (also for wording).

hugovk (approved):
Wording is fine 👍

PR #579, opened 2024-02-24
Fixes #578 - Do not mutate data in place when applying formatters

### LINKED ISSUE #578: Output formatter corrupts original data
version: 3.5.0
python 3.11

I'm trying to export the same dataset several times in different file formats. I've found that specifying a formatter for a column data alters the underlying data. The issue occurs if exporting the same format too.

```python
import tablib

dataset = tablib.Dataset()
dataset.headers = ["text"]
dataset.append(["aAbBcC"])
dataset.add_formatter("text", lambda s: s[2:])

print(dataset.export("csv"))
print(dataset.export("csv"))
```
Output:
```
text
bBcC

text
cC
```

claudep:


hugovk (approved):

claudep:
Thanks!

PR #585, opened 2024-03-22
Update history for 3.6.0

claudep:


claudep:
@hugovk I think it would be time for a new release, if it's ok for you.

hugovk (approved):
Yes, it's been a while! Thanks!

PR #589, opened 2024-04-03
Add optional requirement specifier "html" back

### LINKED ISSUE #588: Removing the html extra dependency should be considered a breaking change
f3ef2e94b40478521b646b58a25dd1b3c5b300f1 introduced a regression by breaking every dependent specifying `tablib[html]`.
Assuming this project does semver (did not find a trace of that in the docs), this should be considered a breaking change unfit for y-releases.
I think the proper way solving this is keeping the "html" optional requirements group and just empty it until the next major release.

mdellweg:
Removing such a specifier should be considered a breaking change and be postponed to the next major release, because `pip install tablib[html]` fails on this. Since there are no actual dependencies to consider, an empty list is put behind that key.

fixes #588

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -26,6 +26,7 @@ Here is a list of past and present much-appreciated contributors:
     Mark Walling
     Mathias Loesch
     Matthew Hegarty
+    Matthias Dellweg
     Mike Waldner
     Peyman Salehi
     Rabin Nankhwa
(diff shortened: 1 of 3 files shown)


hugovk (approved):

hugovk:
Please could you also add this as a bugfix in https://github.com/jazzband/tablib/blob/master/HISTORY.md in this PR, with 3.6.1 and today's date, and I'll make a new release?

hugovk:
Thanks, this has been released: https://pypi.org/project/tablib/3.6.1/

mdellweg:
Thank you so much for pushing it through.

PR #594, opened 2024-07-09
feat: ods: add style to datetime, date and time values 

pascalfree:
This solves #210 but for the ods format.

Let me know if you agree with this approach in general. I will then apply the same for datetime and time types.

Code is inspired by https://stackoverflow.com/a/28577394
I tested it with LibreOffice 7.3 so far.

### DIFF
--- a/src/tablib/formats/_ods.py
+++ b/src/tablib/formats/_ods.py
@@ -5,7 +5,7 @@ import datetime as dt
 import numbers
 from io import BytesIO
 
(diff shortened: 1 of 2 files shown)


pascalfree:
I changed the title and added tests.

The tests currently fail, because the datetime export has a bug that was previously covered by the import: #595

claudep:
Please rebase after merge of #595

claudep (approved):
Thanks a lot!

PR #595, opened 2024-07-10
Fix ods datetime and time export and import

pascalfree:
The ods format was not matching the [spec](https://docs.oasis-open.org/office/OpenDocument/v1.3/os/part3-schema/OpenDocument-v1.3-os-part3-schema.html#__RefHeading__1417680_253892949) (Table 14) on datetime and time:

- datetime must be handled like date: The value goes into the `date-value` attribute instead of the `value` attribute.
- time value must be formatted as a `duration` according to XML spec.

## TODO

- [x] Add backwards compatibility, to import the broken time format as it is exported in the current version?
- [x] Add test for time import
- [x] Add test for datetime export


claudep:
Thanks! I guess I'll wait for the test before approving? You could also add an entry in HISTORY.md

claudep (approved):
Looks good, thanks!

claudep:
Could you still rebase and squash, please?

HISTORY.md
```diff
@@ -1,5 +1,11 @@
 # History
 
+## Dev
```
hugovk, line 3:
```suggestion
## Unreleased
```

hugovk (approved):

PR #598, opened 2024-10-06
Drop support for Python 3.8

hugovk:
Python 3.8 EOL is due tomorrow, when 3.13.0 final is released.

* https://peps.python.org/pep-0569/


claudep (approved):
Thanks!

hugovk:
I think we're due for a release, I can do it later today.

PR #604, opened 2024-11-23
remove redundant check from Dataset.load()

### LINKED ISSUE #603: possible bug/typo in core.py L418 (Dataset.load): reference to global `import_set`
The current code of Dataset.load()  has:
```python
    def load(self, in_stream, format=None, **kwargs):
        ...
        if not import_set:
            raise UnsupportedFormat(f'Format {format} cannot be imported.')
```

I am not sure what is the purpose of this condition (maybe remainder from older code, or maybe should be `fmt.import_set`), but as it is now, since there is no local variable of that name, it actually refers to the module-level function `import_set`.

  This means that the condition checked here will never occur (unless the module's namespace is overwritten in runtime), and it is redundant.

AmitAronovitch:
resolves  #603

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -415,9 +415,6 @@ class Dataset:
         if not hasattr(fmt, 'import_set'):
             raise UnsupportedFormat(f'Format {format} cannot be imported.')
 
-        if not import_set:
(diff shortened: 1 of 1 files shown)


claudep (approved):
Thanks a lot! :heart_eyes:

PR #606, opened 2025-01-19
Add more pre-commit-hooks lint

hugovk:
This adds some extra lint rules to pre-commit that I often use in other projects, and also autofix the whitespace and requirements.txt things it finds.


claudep (approved):
Thanks!

PR #607, opened 2025-01-22
Add support for Python 3.14

hugovk:
Python 3.14 is still in alpha, but I think we can start supporting it already?

* https://peps.python.org/pep-0745/

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -15,7 +15,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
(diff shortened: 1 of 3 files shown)


hugovk:
This repo is pretty quiet and the extra CI time won't make much difference, but I'm fine to leave it until the beta if you prefer?

claudep:
Considering MS is building nuclear powerplants to power its infrastructure, I'd rather not push too much to add more compute power when there is no compelling added value :radioactive: :scream:

PR #608, opened 2025-01-22
Defer imports to improve startup speed

hugovk:
It can be useful to defer imports so that when people import tablib but don't use certain features (for example, if they don't use JSON format) they can avoid the import cost.

This particularly helps people writing tools like CLIs where startup cost can be a problem.

If people are using tablib and other libraries, the total import times can all add up.

We've been doing this also in the stdlib, for example:

* 3.13: https://github.com/python/cpython/issues/109653
* 3.14: https://github.com/python/cpython/issues/118761

Running:

```sh
python3 -X importtime -c 'import tablib' 2> import.log
pip install tuna
tuna import.log
```

# Before: 8 ms

image

# After: 2 ms

image

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -8,9 +8,6 @@
     :license: MIT, see LICENSE for more details.
 """
 
-from copy import copy
-from operator import itemgetter
-
 from .exceptions import (
     HeadersNeeded,
     InvalidDatasetIndex,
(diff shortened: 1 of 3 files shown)


claudep:
I think this makes sense when you are loading known slow external packages. But as far as the standard lib is concerned, I'm seeing this as some "overoptimization". So not very fond of this patch, sorry.

hugovk:
Small improvements multiplied by millions of users and computers could definitely help avoid some of the compute power concerns in https://github.com/jazzband/tablib/pull/607, but okay!

claudep:
I was waiting for such an answer :rofl: You make a point!

Until now, I learned from Python best practices that "embedded" imports were mostly reserved to cases where there was circular imports. If you have some articles about that "conflict" between best practices and optimizations (I guess it's some sort of balance to do), I'm interested.

PR #609, opened 2025-01-22
Update history for 3.8.0

hugovk:


claudep:
What about directly setting the version and date of the release? Do you think it would be cleaner if set in a separate commit?

hugovk:
Good catch, I forgot it. Updated!

claudep (approved):
:+1:

hugovk:
Released! 🚀

Issue #610, opened 2025-01-25
Documentation says python 3.6+, but tablib 3.7.0 dropped support for python 3.8


ebaumgartner:
The "Pythons Supported" section of `docs/intro.rst` says "Python 3.6+ is officially supported."

This seems a bit out of date. Checking the release notes, it looks to me like:

- The last version to support Python 3.6 was tablib 3.1.0 (see #513)
- The last version to support Python 3.7 was tablib 3.4.0 (see #551)
- The last version to support Python 3.8 was tablib 3.6.1 (see #598)

I've integrated tablib into code bases that can't upgrade their Python version and it's a little confusing to have to figure out why the latest version of tablib doesn't work.

A quick fix would be to just update the language to say "Python 3.9+ is officially supported".

Or, if Python 3.6 is still considered officially supported since tablib 3.1.0 supports it, maybe add a couple lines documenting what versions of tablib one should use for older Pythons.

hugovk:
Thanks for the report. 

Python 3.6 is no longer supported and that line is out of date. 

I updated it to 3.9 in https://github.com/jazzband/tablib/pull/607 but we're not going to merge that yet, so let's update it separately.

And on second thoughts, I suggest we remove that line altogether, as we've already forgotten to update it three times (other projects also tend to do the same), and it duplicates the definitive metadata in `pyproject.toml`.

claudep:
+1 to remove that section.

PR #611, opened 2025-01-25
Remove outdated Python support from docs

hugovk:
Fixes https://github.com/jazzband/tablib/issues/610.

Refer to `pyproject.toml` metadata instead.

https://github.com/jazzband/tablib/blob/51e24c2c9f0798df134ae39326a9595a839423f4/pyproject.toml#L18-L31

### DIFF
--- a/docs/intro.rst
+++ b/docs/intro.rst
@@ -51,3 +51,5 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
 THE SOFTWARE.
 
 .. _`The MIT License`: https://opensource.org/licenses/mit-license.php
+
+Now, go :ref:`install Tablib <install>`.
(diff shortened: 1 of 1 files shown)


docs/intro.rst
```diff
@@ -51,12 +51,3 @@ OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
...
 .. _`The MIT License`: https://opensource.org/licenses/mit-license.php
-
-.. _pythonsupport:
-
-Pythons Supported
------------------
-
-Python 3.6+ is officially supported.
-
-Now, go :ref:`install Tablib <install>`.
```
claudep, line 62: Was removal of that last line also intended? Do you think it doesn't bring any value?

claudep: Not sure either. But it looks like most of help pages are redirecting to the next section at the bottom of the page. Putting the link before the licence is doable, but could imply that reading the licence is of no use :smile:. I know, I know, nobody read licences!

hugovk: I added it back, because the other pages also have links at the bottom to the next one.

claudep (approved):
Thanks!

PR #614, opened 2025-02-22
Fix bug in ODS import where headers are not set on Dataset

istride:
Sheets in ODS files saved by LibreOffice can contain 'table-column' nodes before the 'table-row' nodes. Previously, all child nodes of a sheet where enumerated before iterating over them, leading to incorrect counting of skipped rows and failed identification of the headers row.

This change creates a generator that returns 'table-row' nodes only, before they are enumerated and iterated over.


claudep (approved):
Thanks a lot for your contribution in improving the ODS support!

PR #616, opened 2025-03-15
Support 'number-columns-repeated' attribute of ODS cells

istride:
Repeat a cell's value if the 'number-columns-repeated' attribute is set.


src/tablib/formats/_ods.py
```diff
@@ -110,22 +110,33 @@ def import_sheet(cls, dset, sheet, headers=True, skip_lines=0):
 
         dset.title = sheet.getAttribute('name')
 
-        def is_real_cell(cell):
-            return cell.hasChildNodes() or not cell.getAttribute('numbercolumnsrepeated')
+        def repeat(cell):
+            try:
+                n = int(cell.getAttribute("numbercolumnsrepeated"))
+            except Exception:
```
hugovk, line 116: Can we use more specific exceptions here?

src/tablib/formats/_ods.py
```diff
@@ -110,22 +110,33 @@ def import_sheet(cls, dset, sheet, headers=True, skip_lines=0):
...
+            row_vals = [c for cell in row.childNodes for c in repeat(cell)]
             if not row_vals:
                 continue
             if i == skip_lines and headers:
-                dset.headers = row_vals
+                try:
+                    end = row_vals.index("")
+                except ValueError:
+                    end = len(row_vals)
+                dset.headers = row_vals[:end]
```
claudep, line 134: I'm a bit worried here that a column with an empty header cell will cut off data. I suppose having an empty header in the middle of headers is not well tested currently, probably something to improve.

src/tablib/formats/_ods.py
istride: It might be possible just to catch `ValueError` and `AttributeError`, but I really do mean to catch all errors, so I can't recommend it. I think it's safe to assume that if anything goes wrong here, the cell value should not be repeated.

claudep: I would prefer catching more specific exceptions, too. If we get another error, that would mean that something went wrong in an unexpected way, and in that case, it's better to fail loudly than ignoring it.

src/tablib/formats/_ods.py
istride: If I revert this, then `test_ods_import_set_ragged` will fail because the first row of 'ragged.ods' contains 16,380 trailing empty cells. If this constitutes a valid header row, then the assertion in the test would need to be modified to accommodate it. Should I do this instead?

src/tablib/formats/_ods.py
claudep: Thanks!

src/tablib/formats/_ods.py
claudep: Should we use some heuristic to guess the end of the headers, like 5 successive empty headers would mean the header line is over?

istride: @claudep I've decided to follow the convention used in other formats, of accepting the header row as-is. Would you let me know if this ok?

claudep: The problem is that in my experience, most .ods files resulting from an xlsx import (a common use case) will almost always have >16000 rows and also a big number of repeated (empty) rows at the end. This will result in tablib in big data structures mostly filled with empty strings. I would really try to avoid that, even if one could consider this as an ods import bug.

PR #619, opened 2025-04-26
Add sql export format

takhogan:
Hi! I've added .sql to the export formats and added tests

This is useful especially with the records library https://pypi.org/project/records/ which uses this library to create exports.

examples:
```python
import tablib

with open('tests/files/founders.xlsx', 'rb') as f:
    data = tablib.Dataset().load(f, 'xlsx')

sql_output = data.export('sql', table='founders')
print(sql_output) 

"""
INSERT INTO "founders" VALUES ('John', 'Adams', 90);
INSERT INTO "founders" VALUES ('George', 'Washington', 67);
INSERT INTO "founders" VALUES ('Thomas', 'Jefferson', 50);
"""

ds = tablib.Dataset(title='tbl')
ds.headers = ['col_date', 'col_timestamp']
ds.append([
    dt.date(2020, 1, 2),
    dt.datetime(2020, 1, 2, 3, 4, 5)
])
sql = ds.export('sql')
print(sql)

"""
INSERT INTO "tbl" VALUES (DATE '2020-01-02', TIMESTAMP '2020-01-02 03:04:05');
"""
```


docs/formats.rst
```diff
@@ -300,3 +300,32 @@ This format is optional, install Tablib with ``pip install "tablib[yaml]"`` to
 make the format available.
 
 .. _YAML: https://yaml.org
+
+sql
+===
+
```
claudep, line 306: You might include a `.. versionadded:: 3.9.0` directive.

claudep (approved):
Thanks, I think this is a nice addition to this lib.

src/tablib/formats/_sql.py
```diff
@@ -45,18 +40,26 @@ def _render_literal(value):
...
     @classmethod
-    def export_set(cls, dataset, table=None):
+    def export_set(cls, dataset, table=None, columns=None, commit=False):
         """
         Return SQL INSERT statements for Dataset rows.
         :param table: optional table name; defaults to dataset.title or 'data'
         """
         tbl = table or getattr(dataset, 'title', None) or 'EXPORT_TABLE'
-        tbl_ident = cls._quote_ident(str(tbl))
+        tbl_ident = str(tbl)
```
claudep, line 49: What is your rationale for not quoting table/column names? Typically, I think that for example on PostgreSQL uppercase names should be quoted.

takhogan: Hi when the table name is in quotes then you're not able to reference schema names like schema_name.table_name so "schema_name.table_name" doesn't work. Seems like you can have syntaxes like "Schema Name"."table name" as well. I think the user should just pass the quotes into the title if they require it instead of having it as default

claudep: Maybe `_quote_ident` could smartly detect the "schema_name.table_name" case? In any case, having the default EXPORT_TABLE unquoted looks like it will hurt many.

claudep:
@hugovk Could you quickly make a final review?

src/tablib/formats/_sql.py
```diff
@@ -0,0 +1,72 @@
...
+    def _render_literal(value):
+        """Render a Python value as an SQL literal."""
+        if value is None:
+            return 'NULL'
+        if isinstance(value, bool):
+            return 'TRUE' if value else 'FALSE'
+        if isinstance(value, int) and not isinstance(value, bool):
+            return str(value)
+        if isinstance(value, decimal.Decimal):
+            return str(value)
```
hugovk, line 23: No need to repeat the bool check, and can combine two of these:
```suggestion
        if isinstance(value, bool):
            return 'TRUE' if value else 'FALSE'
        if isinstance(value, (int, decimal.Decimal):
            return str(value)
```

src/tablib/formats/_sql.py
```diff
@@ -0,0 +1,72 @@
...
+        if isinstance(value, float):
+            # Represent finite floats; non-finite as NULL
+            try:
+                if value == value and value not in (float('inf'), -float('inf')):
+                    return repr(value)
+            except Exception:
+                pass
+            return 'NULL'
```
hugovk, line 31: What might trigger this? Can we have a more specific exception here? Can it be covered by tests?

hugovk (commented):
Looks good, a couple of suggestions.

src/tablib/formats/_sql.py
takhogan: Honestly, I'm not sure. It would have to be something obscure like a class that subclasses the float type and throws an exception in it's repr method

hugovk (approved):

claudep:
Thanks a lot for this contribution :heart_eyes:

PR #620, opened 2025-05-15
Improve comprehensions

hugovk:
These are more efficient. Identified with the flake8-comprehensions (C4) tool in Ruff:

https://docs.astral.sh/ruff/rules/#flake8-comprehensions-c4

We could also add the flake8-comprehensions plugin to our config, or even the Ruff equivalent for more efficient linting as well?

### DIFF
--- a/src/tablib/_vendor/dbfpy/record.py
+++ b/src/tablib/_vendor/dbfpy/record.py
@@ -242,7 +242,7 @@ class DbfRecord:
             real values stored in this object.
 
         """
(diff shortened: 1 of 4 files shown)


claudep (approved):
:+1:

hugovk:
> We could also add the flake8-comprehensions plugin to our config, or even the Ruff equivalent for more efficient linting as well?

What do you think about these options? 

Would make sure future code also follows these patterns.

claudep:
I'm fine with either one.

PR #621, opened 2025-05-16
Replace Flake8 with Ruff

hugovk:
It found a few extra pyupgrade things to fix.

Time to run the whole thing reduced from 2.3 to 1.2 seconds.


claudep (approved):
:+1:

Issue #622, opened 2025-05-20
Simple csv file can't be parsed correctly because of Sniffer() parameters


lap1nou:
Greetings,

First of all, thank you for the work you are doing on this library.

I have the following CSV file, where the delimiter is a colon `:` :
```csv
test:test2
value1:value2
```

I'm now using the `tablib` library to import it:
```python
tmp = open('test.csv', 'r')
tmp2 = tablib.Dataset().load(tmp, format='csv')
```

But the file is wrongly parsed, because the `csv` library did not sniffed the delimiter correctly:
```
test:test2   
-------------
value1:value2
```

I saw that you are passing `2048 bytes` of samples to be sniffed (https://github.com/jazzband/tablib/blob/4bbb35a4103fea2e091c15314f0a1c53fc9613cb/src/tablib/formats/_csv.py#L58), I now tested to manually sniff the CSV using only the first line:
```python
dialect = csv.Sniffer().sniff(tmp.readline())
```

The delimiter is now indeed recognized as a colon `:`.

My question is: shouldn't you pass the first line to sniff only (using `readline()`) ? Or maybe fallback on the actual behaviour only if the headers is not present in the file ?

Related issue: #502

Regards

claudep:
I think we are open to try better heuristics to detect delimiters, as long as current tests still pass, of course.

lap1nou:
@claudep After analysis I think I found a bug.

There: https://github.com/jazzband/tablib/blob/8dcb87f69d2f6d501a511210387df233099b2957/src/tablib/formats/_csv.py#L41, you are setting a default value if the `delimiter` parameter is not set, however if someone decide to pass a `dialect` in the `kwargs`, the `delimiter` parameter won't be present, and so the default value you are setting will take precedence over the dialect passed by the user, here is an example using the following CSV:

```csv
test:test2
value1:value2
```

```python
dialect = csv.Sniffer().sniff(csv_file.readline())
csv_file.seek(0)

data = tablib.Dataset().load(
                csv_file, format='csv', dialect=dialect
)
```

Since no delimiter are passed, tablib will choose the deault one `,`, but we actually have the delimiter `:` in the dialect, so the default delimiter will take precedence over the dialect one and it will fail to correctly import the CSV.

Do you want me to submit a PR to fix this bug ?

claudep:
Absolutely, if you can show the bug with a regression test, any PR is welcome!

PR #623, opened 2025-05-20
Modified CSV Sniffer sample

lap1nou:
Greetings,

In this PR I modified the CSV Sniffer sample, reading the first line instead of 2048 bytes.

Related to #622.

The idea come from the answer of this SO question: https://stackoverflow.com/questions/35756682/getting-csv-sniffer-to-work-with-quoted-values

All current tests are passing.

Regards.

### DIFF
--- a/src/tablib/formats/_csv.py
+++ b/src/tablib/formats/_csv.py
@@ -55,7 +55,7 @@ class CSVFormat:
     def detect(cls, stream, delimiter=None):
         """Returns True if given stream is valid CSV."""
         try:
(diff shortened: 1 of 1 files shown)


claudep:
Thanks! Please add a regression test matching your use case (should fail before the patch, pass after).

lap1nou:
Hey @claudep,

I think I misunderstood the role of the sniffing in the library, it is just used to detect if it's a CSV or not, but not to detect the dialect used by the file as I understood it, so my patch doesn't actually do anything.

I will just close the PR / issue, sorry for that.

PR #624, opened 2025-06-26
feat: support column_width in xlsx format for databook exports

npilon:
Extend work from #516 to support exports of Databook objects to XLSX with column widths, as already supported for Dataset. I read the new tests added in that PR and did not see any that would have covered the `export_set` lines that I've copied.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -31,6 +31,7 @@ Here is a list of past and present much-appreciated contributors:
     Matthew Hegarty
     Matthias Dellweg
     Mike Waldner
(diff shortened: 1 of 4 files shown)


claudep:
Thanks for your contribution. I guess it should still be welcome to add a test with the `column_width` parameter to a Databook export. And probably the docs needs a small update too.

npilon:
@claudep is there anything else needed from me on this change?

claudep (approved):
Looks good, thanks!

Issue #303, opened 2017-08-29
Databook sheet title must be below 31 characters for XLSX


mpasternak:
Hi,

for XLSX format, databook sheet title must be below 32 characters. If it is longer, you get a warning about a corrupted file when trying to open such file in Microsoft Excel (the original MS Excel from MS Office package).

Attached is out file and a minimal non-working code example. 

```
import tablib

parent_data = tablib.Databook()

data = tablib.Dataset()
# collection of names
names = ['Kenneth Reitz', 'Bessie Monke']

for name in names:
    # split name appropriately
    fname, lname = name.split()

    # add names to Dataset
    data.append([fname, lname])

data.title = "1" * 32
parent_data.add_sheet(data)

x = open("test.xlsx", "wb")
x.write(parent_data.xlsx)
x.close()
```

This is the out file that I got on Python 3.6 and tablib 0.11.5. 

[test.xlsx](https://github.com/kennethreitz/tablib/files/1258761/test.xlsx)

When I change the dataset title length to below 31 characters, everything goes back to normal.

hugovk:
With Python 3.7.4, Tablib 0.14.0 and openpyxl 3.0.0:

```console
$ python3 303.py
/usr/local/lib/python3.7/site-packages/openpyxl/workbook/child.py:99: UserWarning: Title is more than 31 characters. Some applications may not be able to read the file
  warnings.warn("Title is more than 31 characters. Some applications may not be able to read the file")
```

Opening the file:

![image](https://user-images.githubusercontent.com/1324225/67300243-59f6f600-f4f6-11e9-9112-16d33574e478.png)

Clicking Yes:

![image](https://user-images.githubusercontent.com/1324225/67300366-8448b380-f4f6-11e9-9bd1-6bd4e7d367b3.png)

Clicking View, this file opens in a text editor:
 /private/var/folders/kt/j77sf4_n6fnbx6pg199rbx700000gn/T/com.microsoft.Excel/Repair Result to test0.xml

```xml
<?xml version="1.0" encoding="UTF-8" standalone="yes"?>
[...]

PR #627, opened 2025-08-26
Add dataset title adjustment for xls format

TheSuperiorStanislav:


claudep (commented):
Thanks for the patch! You should mention in the commit description that it fixes #303.

I'm not convinced that it's worth allowing customizing the replacement character. Maybe just having it as a class property would be enough. @hugovk, thoughts?

claudep (approved):

PR #629, opened 2025-10-04
Drop support for Python 3.9

hugovk:
It's almost end-of-life:

* https://peps.python.org/pep-0596/

Also bump GitHub Actions.


claudep:
Thanks, but what about releasing 3.9.0 before removing Python 3.9 support?

claudep:
Awesome, thanks!

claudep (approved):
Thanks, looks good!

PR #631, opened 2025-10-13
Use Release Drafter to make releasing easier

hugovk:
Before making the next release, I'd like to propose some more automation using [Release Drafter](https://github.com/marketplace/actions/release-drafter), to make releasing easier and take out some of the drudgery. I'm using this in more than a dozen other projects.

The new workflow would be something like:

Before merging PRs:

* Add one of the following labels:

  - changelog: Added
  - changelog: Changed
  - changelog: Deprecated
  - changelog: Fixed
  - changelog: Removed
  - changelog: Security
  - changelog: skip

* Check the PR title is suitable to be shown in the release page

After merge:

* The action runs and it creates a draft release under https://github.com/jazzband/tablib/releases. Only committers can see this, until published.
* It groups the changes under those headers.
* And because we use SemVer, it also pre-fills the next version number. So for example, if there's just "Changelog: Fixed", it's a micro bump; if "Changelog: Added", it's minor, and if "Changelog: Removed", it's major.
* "Changelog: skip" is for tests and things that aren't of interest to the end user and don't need adding to the changelog.

At release:

* When it's time to release, edit the draft and adjust the text if needed, then publish like before

Benefits:

* We don't need to manually trawl through the PRs from last time and manually update https://github.com/jazzband/tablib/blob/master/HISTORY.md
* You pick a title for the changelog entry at merge time, when the PR is fresh in your mind
* Likewise, you pick SemVer levels at merge time
* Pre-fills the release entry according to SemVer
* Less for us to fill in
* Can check the draft at any time to see how much is ready for release

Downsides:

* We don't update https://github.com/jazzband/tablib/blob/master/HISTORY.md any more, and the changelog isn't in Git (but I've had zero problems about this in other projects and it makes things so much easier)
* More automation to maintain (but I'm maintaining it in other projects so it's the same thing)

What do you think?

I'll add the labels if it's okay to go ahead.

### DIFF
--- a/.github/release-drafter.yml
+++ b/.github/release-drafter.yml
@@ -1,0 +1,48 @@
+name-template: "v$RESOLVED_VERSION"
+tag-template: "v$RESOLVED_VERSION"
+
+categories:
+  - title: "Added"
+    labels:
+      - "changelog: Added"
+      - "enhancement"
+  - title: "Changed"
+    label: "changelog: Changed"
+  - title: "Deprecated"
+    label: "changelog: Deprecated"
+  - title: "Removed"
+    label: "changelog: Removed"
+  - title: "Fixed"
+    labels:
+      - "changelog: Fixed"
+      - "bug"
+  - title: "Security"
+    label: "changelog: Security"
+
+exclude-labels:
+  - "changelog: skip"
+
+autolabeler:
+  - label: "changelog: skip"
+    branch:
+      - "/pre-commit-ci-update-config/"
+
+template: |
+  $CHANGES
+
+version-resolver:
+  major:
+    labels:
+      - "changelog: Removed"
+  minor:
+    labels:
+      - "changelog: Added"
+      - "changelog: Changed"
+      - "changelog: Deprecated"
+      - "enhancement"
+
+  patch:
+    labels:
+      - "changelog: Fixed"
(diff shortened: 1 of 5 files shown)


hugovk:
> Please go ahead, we'll see in practice what's the result will be. I'm fully trusting you 🤩

Thanks, let's see how it goes, and happy to make adjustments in the future!

https://github.com/jazzband/tablib/actions/runs/18468250067/job/52615477509?pr=631

> stack: 'Error: Configuration file .github/release-drafter.yml is not found. The configuration file must reside in your default branch.\n' +

This error should go away after merge.

> At least 1 approving review is required by reviewers with write access.

Please could you approve it?

claudep (approved):

hugovk:
Right, then I clicked the pencil icon, checked if anything needed editing (I decided it was okay), and then clicked Publish.

That kicked off GitHub Actions to do the release: https://github.com/jazzband/tablib/actions/runs/18538580208

Which then did the usual Jazzband thing, and a couple of clicks later, it's now up on PyPI!

https://pypi.org/project/tablib/3.9.0/

hugovk:
And after release, we can see new things appearing in the next draft:

image

https://github.com/jazzband/tablib/releases

PR #637, opened 2026-03-27
Prevent XLS precision loss for large integers

### LINKED ISSUE #197: Wrong long int value exported
I have the code:

    import tablib

    data = tablib.Dataset()
    data.headers = ['test_id']

    a = 568883383628111872
    data.append([a])
    with open('test.xls', 'wb') as f:
        f.write(data.xls)

Which produces one `xls` file with id as `568883383628112000`.

ViceSaber:
## Summary
- prevent precision loss when exporting large integers to XLS format
- integers exceeding XLS float precision (2^53 - 1) are now stored as text strings

## Problem
XLS (Excel 97-2003) stores numeric values as 64-bit IEEE 754 floats with a 53-bit mantissa. Integers larger than 9,007,199,254,740,991 (2^53 - 1) lose precision when stored as numbers. For example, `568883383628111872` becomes `568883383628112000`.

## Fix
In `dset_sheet`, detect integers that exceed the float precision limit and convert them to strings before writing. Small integers continue to be stored as numbers for normal behavior.

## Testing
Verified that:
- Large integers (e.g., `568883383628111872`) are stored as text (cell type 1) and read back with full precision
- Small integers (e.g., `42`) continue to be stored as numbers (cell type 2)
- Boolean values are unaffected (not coerced to string)

Fixes #197

### DIFF
--- a/src/tablib/formats/_xls.py
+++ b/src/tablib/formats/_xls.py
@@ -150,6 +150,12 @@ class XLSFormat:
         for i, row in enumerate(_package):
             for j, col in enumerate(row):
 
+                # Convert large ints to str to avoid XLS float precision loss
+                # (XLS uses 64-bit floats: precision is limited to ~15 digits)
+                if isinstance(col, int) and not isinstance(col, bool):
+                    if abs(col) > 9007199254740991:  # 2^53 - 1
+                        col = str(col)
+
                 # bold headers
                 if (i == 0) and dataset.headers:
                     ws.write(i, j, col, bold)


hugovk:
Thanks for the PR, please add tests.

I've just merged a fix for the lint teyit failure, please also rebase/merge `main` to include that.

claudep:
Another approach would have been to document that the XLS format is not adequate to store large integers. Is that format not on a disappearing trend?

hugovk:
I'm fine with either approach.

This fix seems simple enough, but I expect XLS use is very low, and outputting XLS even lower.

A docs change is safer, and we've not exactly had a huge number of requests for this.

PR #639, opened 2026-04-17
Auto-detect CSV delimiter via `Sniffer` in `import_set` and `detect()`

### LINKED ISSUE #622: Simple csv file can't be parsed correctly because of Sniffer() parameters

terminalchai:
Fixes #622

## Problem

`tablib.Dataset().load(data, format='csv')` always uses `,` as the delimiter, even when the file uses `;`, `:`, `|`, or other separators. This is because `import_set` calls `kwargs.setdefault('delimiter', cls.DEFAULT_DELIMITER)` without ever sniffing the actual content.

A secondary issue: `CSVFormat.detect()` passed `delimiters=delimiter or cls.DEFAULT_DELIMITER` to `csv.Sniffer().sniff()`, which restricted sniffing to comma-only. This meant non-comma CSV files were never recognised during format auto-detection.

## Fix

**import_set**: when no `delimiter` kwarg is provided, read up to 2048 bytes as a sample, call `csv.Sniffer().sniff(sample)` with no delimiter restriction, and use the detected delimiter. The stream is then seeked back to 0 before the actual read. A guard rejects alphabetic/digit detections (the Sniffer occasionally misidentifies a letter on very short or ambiguous samples). Falls back to `DEFAULT_DELIMITER` on any `csv.Error`.

**detect()**: when no explicit delimiter is given, uses a candidate string of common non-tab separators (',;:|') for CSVFormat, and the format's `DEFAULT_DELIMITER` for subclasses (e.g. TSVFormat uses tab). This keeps tab-delimited files out of CSV auto-detection while still recognising `;`, `:`, and `|` separated files.

Explicit `delimiter` kwargs are fully respected and take precedence.

## Tests

- Updated `test_csv_formatter_support_kwargs`: old assertion documented the broken behaviour (1 header instead of 3); updated to assert correct parsing.
- Added `test_csv_import_set_auto_detect_delimiter`: colon- and pipe-delimited CSV auto-detection.
- Added `test_csv_detect_non_comma_delimiter`: `detect()` correctly recognises `:`, `|`, and `;` separated files.

All 155 tests pass.

### DIFF
--- a/src/tablib/formats/_csv.py
+++ b/src/tablib/formats/_csv.py
@@ -38,7 +38,27 @@ class CSVFormat:
 
         dset.wipe()
 
-        kwargs.setdefault('delimiter', cls.DEFAULT_DELIMITER)
+        if 'delimiter' not in kwargs:
+            # Auto-detect the delimiter using the CSV Sniffer.
+            # Reading a sample and then seeking back to the start lets the
+            # Sniffer work on real data while keeping the stream intact for
+            # the subsequent csv.reader call.  Crucially, we do NOT restrict
+            # sniffing to a single delimiter so that files using non-comma
+            # separators (e.g. ';', ':', '|') are imported correctly.  See #622.
+            try:
+                sample = in_stream.read(2048)
+                dialect = csv.Sniffer().sniff(sample)
+                # Only accept non-word-character delimiters (true separators
+                # such as ',', ';', '|', ':', '\t').  The Sniffer occasionally
+                # misidentifies a letter as a delimiter when the sample is short
+                # or ambiguous; falling back to the default in that case is safer.
+                if dialect.delimiter.isalpha() or dialect.delimiter.isdigit():
+                    raise csv.Error("unlikely delimiter detected, falling back")
+                kwargs['delimiter'] = dialect.delimiter
+            except csv.Error:
+                kwargs['delimiter'] = cls.DEFAULT_DELIMITER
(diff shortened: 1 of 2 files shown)


ClemenceAlain:
[praise](https://pullpo.io/cc?l=praise) 

Nice job ! We have some ugly hack to test multiple delimiters so I came to this repo only to see that your proposition already answers our issue. I don't have the rights to review but I tested and can tell it works fine :+1:

Issue #524, opened 2022-08-15
KeyError reading yaml file


roderick-liu:
yaml file:
```yaml
mylist:
  - 1
  - 2
  - 3
```

```
with open("config/aa.yaml","rb")as f:
     data.load(f,'yaml')

---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
Input In [21], in <module>
      1 with open("config/aa.yaml","rb")as f:
----> 2      data.load(f,'yaml')

File ~/anaconda3/envs/py3/lib/python3.9/site-packages/tablib/core.py:414, in Dataset.load(self, in_stream, format, **kwargs)
    411 if not import_set:
    412     raise UnsupportedFormat(f'Format {format} cannot be imported.')
--> 414 fmt.import_set(self, stream, **kwargs)
    415 return self

File ~/anaconda3/envs/py3/lib/python3.9/site-packages/tablib/formats/_yaml.py:32, in YAMLFormat.import_set(cls, dset, in_stream)
     29 """Returns dataset from YAML stream."""
     31 dset.wipe()
---> 32 dset.dict = yaml.safe_load(in_stream)

File ~/anaconda3/envs/py3/lib/python3.9/site-packages/tablib/core.py:341, in Dataset._set_dict(self, pickle)
    338     return
    340 # if list of rows
--> 341 if isinstance(pickle[0], list):
    342     self.wipe()
    343     for row in pickle:

KeyError: 0
```

hugovk:
Please can you give a complete, minimal, self-contained Python script which reproduces your problem?

Tips here: https://stackoverflow.com/help/minimal-reproducible-example

claudep:
tablib expects a yaml content structured with headers, like:
```
- {first_name: John, gpa: 90, last_name: Adams}
- {first_name: George, gpa: 67, last_name: Washington}
```
The docs could be more explicit and the error message in case of failure, too.

MikiVanousek:
I have the same issue. Minimal code for reproduction:
`props.yaml`:
```yaml
title: Voice of Miki Vanoušek
```
And the python code:
```py
import tablib
with open('props.yaml', 'r') as f:
    podcast_properties = tablib.Dataset().load(f)
```

PR #641, opened 2026-05-20
Remove `#yaml` tag from generic error message URL

mgrdcm:
This error isn't `yaml`-specific.  Looks like this got committed (without PR?) as part of addressing https://github.com/jazzband/tablib/issues/524 and perhaps bypassed PR/review: https://github.com/jazzband/tablib/commit/0a3511fb9789b2e9f0d58739fccffde8db9f2635#diff-026edabea10cc8f56b3dc11a795a25d0bdc86ae602fc6649cd30730b5624ba78R336-L353

Possible there are other related changes around this code that should be made too, but seemed like fixing the misleading error message might help a little.  Here's another later related commit:  https://github.com/jazzband/tablib/commit/b2be12e5b647111d8e0f24c8bab646a97a3d68f2

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -338,7 +338,7 @@ class Dataset:
         """
         error_details = (
             "Please check format documentation "
-            "https://tablib.readthedocs.io/en/stable/formats.html#yaml"
+            "https://tablib.readthedocs.io/en/stable/formats.html"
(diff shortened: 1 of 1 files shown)


claudep (approved):
Thanks for the fix, looks good!

PR #643, opened 2026-05-29
Preserve falsy values (0, False) in jira and latex exports

koriyoshi2041:
The JIRA and LaTeX serializers decide whether a cell is empty with a plain truthiness test:

```python
# _jira.py
delimiter.join([str(item) if item else " " for item in row])
# _latex.py
new_row = [cls._escape_tex_reserved_symbols(str(item)) if item else "" for item in row]
```

So `0`, `0.0`, and `False` — all valid data — are treated as missing and exported as blank cells, silently losing data. For example `tablib.Dataset([0, False, "c"]).jira` produced `| | |c|` instead of `|0|False|c|`.

The HTML format already does this correctly with `str(item) if item is not None else ""`, so "blank only for missing values" is the intended contract. This changes both serializers to test `item not in (None, "")`, which keeps the existing `None`/empty-string behaviour (still rendered blank) while preserving real falsy values.

Added `test_jira_export_falsy_values` and `test_latex_export_falsy_values`; the existing `*_none_*` tests still pass. `pytest -k "jira or latex"` is green (12 passed).

### DIFF
--- a/src/tablib/formats/_jira.py
+++ b/src/tablib/formats/_jira.py
@@ -2,6 +2,7 @@
 
    Generates a Jira table from the dataset.
 """
+from tablib.utils import is_empty_cell
 
 
 class JIRAFormat:
@@ -35,16 +36,6 @@ class JIRAFormat:
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([' ' if cls._is_empty_value(item) else str(item) for item in row]),
+            delimiter.join([' ' if is_empty_cell(item) else str(item) for item in row]),
             delimiter
         )
-
-    @classmethod
-    def _is_empty_value(cls, item):
-        if item is None:
-            return True
-        if isinstance(item, (str, bytes, bytearray)):
-            return len(item) == 0
-        if isinstance(item, (list, tuple, dict, set, frozenset)):
-            return len(item) == 0
-        return False
--- a/src/tablib/formats/_jira.py
+++ b/src/tablib/formats/_jira.py
@@ -35,6 +35,6 @@ class JIRAFormat:
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([str(item) if item else ' ' for item in row]),
+            delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
             delimiter
         )
--- a/src/tablib/formats/_jira.py
+++ b/src/tablib/formats/_jira.py
@@ -35,6 +35,16 @@ class JIRAFormat:
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
+            delimiter.join([' ' if cls._is_empty_value(item) else str(item) for item in row]),
             delimiter
         )
+
+    @classmethod
+    def _is_empty_value(cls, item):
+        if item is None:
+            return True
+        if isinstance(item, (str, bytes, bytearray)):
+            return len(item) == 0
+        if isinstance(item, (list, tuple, dict, set, frozenset)):
+            return len(item) == 0
+        return False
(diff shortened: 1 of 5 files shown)


hugovk (commented):
This also changes the display of other falsy things.

With this script:

```python
import tablib

# The fix itself: 0 and False are real values.
print(repr(tablib.Dataset([0, False, 'c']).jira))

# Finding 1: change is broader than 0/False — falsy containers/bytes now
# render their reprs instead of blank cells (jira and latex alike).
print(repr(tablib.Dataset([[], {}, 'c']).jira))

# Finding 2: b'' hits a bytes-vs-str comparison in `item not in (None, '')`
# — raises BytesWarning under -bb (master rendered a blank cell).
print(repr(tablib.Dataset([b'', 'c']).jira))
```

`main`:

```
'| | |c|'
'| | |c|'
'| |c|'
```
PR:

```
'|0|False|c|'
'|[]|{}|c|'
"|b''|c|"
```

And if we run with bytes warnings: `python -b repro.py`:

`main`:

```
'| | |c|'
'| | |c|'
'| |c|'
```

PR:

```
'|0|False|c|'
'|[]|{}|c|'
/Users/hugo/github/tablib/src/tablib/formats/_jira.py:38: BytesWarning: Comparison between bytes and string
  delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
/Users/hugo/github/tablib/src/tablib/formats/_jira.py:38: BytesWarning: str() on a bytes instance
  delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
"|b''|c|"
```

src/tablib/formats/_latex.py
```diff
@@ -117,7 +117,7 @@ def _serialize_row(cls, row):
         :param row: single dataset row
         """
 
-        new_row = [cls._escape_tex_reserved_symbols(str(item)) if item else ''
+        new_row = [cls._escape_tex_reserved_symbols(str(item)) if item not in (None, '') else ''
```
hugovk: We can simplify this:
```suggestion
        new_row = ['' if item is None else cls._escape_tex_reserved_symbols(str(item))
```

src/tablib/formats/_jira.py
```diff
@@ -35,6 +35,6 @@ def _get_header(cls, headers):
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([str(item) if item else ' ' for item in row]),
+            delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
```
hugovk:
```suggestion
            delimiter.join([' ' if item in (None, '') else str(item) for item in row]),
```

src/tablib/formats/_jira.py
```diff
@@ -35,6 +35,16 @@ def _get_header(cls, headers):
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
+            delimiter.join([' ' if cls._is_empty_value(item) else str(item) for item in row]),
             delimiter
         )
+
+    @classmethod
+    def _is_empty_value(cls, item):
```
hugovk: Move this to utils.py and deduplicate.

src/tablib/formats/_jira.py
```diff
@@ -35,6 +35,16 @@ def _get_header(cls, headers):
...
+        if item is None:
+            return True
+        if isinstance(item, (str, bytes, bytearray)):
+            return len(item) == 0
+        if isinstance(item, (list, tuple, dict, set, frozenset)):
+            return len(item) == 0
```
hugovk: This could be a single:
```python
if isinstance(item, (str, bytes, bytearray, list, tuple, dict, set, frozenset)) and len(item) == 0:
    return True
````

File: src/tablib/formats/_jira.py [hunk 5]
```diff
@@ -35,6 +35,16 @@ def _get_header(cls, headers):
     def _serialize_row(cls, row, delimiter='|'):
         return '{}{}{}'.format(
             delimiter,
-            delimiter.join([str(item) if item not in (None, '') else ' ' for item in row]),
+            delimiter.join([' ' if cls._is_empty_value(item) else str(item) for item in row]),
             delimiter
         )
+
+    @classmethod
```
hugovk: `cls` is never used, can remove `@classmethod` and `cls` at the call site.

PR #644, opened 2026-06-13
Preserve column order in YAML export (#287)

### LINKED ISSUE #287: library re-orders columns when exporting content in YAML
Hello,

I've been using tablib indirectly through records to produce some Excel reports from a database, and I've noticed that while I add my content in a specific order, when exporting, the order of the columns is not the same as I added.

Looking at the documentation of tablib -- if the examples are not merely illustrative -- this should be noticeable, as for instance, when exporting to JSON, YAML, and CSV, the output "columns" are re-ordered differently for each format:

```python
>>> print data.json
[
  {
    "last_name": "Adams",
    "age": 90,
    "first_name": "John"
  },
  {
    "last_name": "Ford",
    "age": 83,
    "first_name": "Henry"
  }
]

>>> print data.yaml
- {age: 90, first_name: John, last_name: Adams}
- {age: 83, first_name: Henry, last_name: Ford}

>>> print data.csv
first_name,last_name,age
John,Adams,90
Henry,Ford,83
```

I'd appreciate if this could be fixed in such way that output would always have the same structure as what was inputted.

Thanks

vineethsaivs:
Fixes #287.

`yaml.safe_dump` sorts mapping keys alphabetically by default, so YAML export reordered columns inconsistently with the CSV and JSON exports (which keep the dataset's header order):

```python
import tablib
d = tablib.Dataset()
d.headers = ["name", "id", "zebra", "apple"]
d.append(["x", 1, "z", "a"])

d.csv.splitlines()[0]   # 'name,id,zebra,apple'
d.yaml                  # before: '- {apple: a, id: 1, name: x, zebra: z}'
                        # after:  '- {name: x, id: 1, zebra: z, apple: a}'
```

This passes `sort_keys=False` when dumping the dataset and the databook, so YAML keeps the original column order (matching CSV/JSON).

`test_yaml_export` is updated to expect header order, and a focused regression test (`test_yaml_export_preserves_column_order`) is added; it fails on `master` (alphabetical) and passes with this change.

---
*Disclosure: prepared with AI assistance; reviewed by me and I can explain every line.*

### DIFF
--- a/tests/test_tablib.py
+++ b/tests/test_tablib.py
@@ -1709,13 +1709,21 @@ class YAMLTests(BaseTestCase):
 
         self.founders.append(('名字', '李', 60))
         expected = """\
-- {first_name: John, gpa: 90, last_name: Adams}
-- {first_name: George, gpa: 67, last_name: Washington}
-- {first_name: Thomas, gpa: 50, last_name: Jefferson}
-- {first_name: 名字, gpa: 60, last_name: 李}
+- {first_name: John, last_name: Adams, gpa: 90}
+- {first_name: George, last_name: Washington, gpa: 67}
+- {first_name: Thomas, last_name: Jefferson, gpa: 50}
+- {first_name: 名字, last_name: 李, gpa: 60}
 """
         output = self.founders.yaml
         self.assertEqual(output, expected)
+
+    def test_yaml_export_preserves_column_order(self):
+        """YAML export must keep the dataset's column order instead of sorting the keys
+        alphabetically."""
+        data = tablib.Dataset()
+        data.headers = ('name', 'id', 'zebra', 'apple')
+        data.append(('x', 1, 'z', 'a'))
+        self.assertEqual(data.yaml.strip(), '- {name: x, id: 1, zebra: z, apple: a}')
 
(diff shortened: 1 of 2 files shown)


tests/test_tablib.py
```diff
@@ -1709,14 +1709,22 @@ def test_yaml_export(self):
...
+- {first_name: John, last_name: Adams, gpa: 90}
+- {first_name: George, last_name: Washington, gpa: 67}
+- {first_name: Thomas, last_name: Jefferson, gpa: 50}
+- {first_name: 名字, last_name: 李, gpa: 60}
 """
         output = self.founders.yaml
         self.assertEqual(output, expected)
 
+    def test_yaml_export_preserves_column_order(self):
+        """Regression test for #287: YAML export must keep the dataset's column order
```
claudep: I think you could remove the "Regression test for #287" part, the rest of the patch looks totally good. Thanks!

vineethsaivs:
Thanks for the review! I've removed the "Regression test for #287" wording from the docstring; the test now reads as a plain description of the expected behavior.

claudep (approved):
Thanks!

PR #645, opened 2026-06-16
Use lazy imports on Python 3.15 to improve startup speed

hugovk:
Replaces and closes https://github.com/jazzband/tablib/pull/608.

The upcoming Python 3.15 (now in beta) introduces a new `lazy` keyword for imports:

```python
lazy import json
```

With this, the import doesn't actually happen until the first use:

```python
json.dumps(thing)
```

So there's no need for the deferred method used in #608.

This is new syntax in 3.15, if you use it on an older version you get a syntax error. But there's a transitional mechanism: if you declare the imports in `__lazy_modules__`, they'll be treated as if they're lazy in 3.15+, and ignored for older versions.

* https://docs.python.org/3.15/whatsnew/3.15.html#whatsnew315-lazy-imports
* https://peps.python.org/pep-0810/

Running with Python 3.15:

```sh
python3 -X importtime -c 'import tablib' 2> import.log
pip install tuna
tuna import.log
```

# Before: 8 ms

image

# After: ~0 ms

image

Even better than https://github.com/jazzband/tablib/pull/608!


claudep:
Interesting, really! I wonder however what's the cost of "hard" importing the Python's own modules, like io, decimal, operator. Aren't essentially external modules that are costly to import?

hugovk:
The light blue boxes in the top chart are stdlib modules, and we can see many of them are contributing real cost to the import tree.

As linked in https://github.com/jazzband/tablib/pull/608, we've been deferring imports inside the stdlib of other stdlib modules to help reduce this cost, and in 3.15+ using the `lazy` keyword instead (for example, https://github.com/python/cpython/pull/147992, https://github.com/python/cpython/pull/149318, https://github.com/python/cpython/pull/149416, https://github.com/python/cpython/pull/146327). `re` is a common target.

The stdlib has some benefit: they're usually byte-compiled to pyc at install time, but that's a cheap one-time cost for external modules, and doesn't affect costly execution time. Some modules are always imported as part of `site` (but that can be disabled). Some bootstrap modules are "frozen" into the binary to avoid file lookup.

Looking the top chart again, most of the lower edge underneath tablib is light blue, so most of the import time comes from importing stdlib.

claudep (approved):
Thanks for the detailed explanations 😍

PR #646, opened 2026-06-16
Bump Read the Docs versions

hugovk:
>  Config file validation error Config validation error in build.os. Expected one of (ubuntu-22.04, ubuntu-24.04, ubuntu-26.04, ubuntu-lts-latest), got type str (ubuntu-20.04). Double check the type of the value. A string may be required (e.g. "3.10" instead of 3.10)

https://app.readthedocs.org/projects/tablib/builds/33169746/

They've dropped Ubuntu 22.04, let's try bumping to 26.04, and also bump Python from 3.10 to 3.14 at the same time.


claudep (approved):
👍

Issue #226, opened 2016-02-24
Support importing ragged CSV text into a Dataset


klothe:
Right now Dataset._validate prevents importing data where not all rows have the same width. This is not possible in some formats but is common in CSV files. Example:

``` python
from tablib import Dataset
from tablib.formats import csv
ds = Dataset()
csv.import_set(ds, "a,b,c\nx,y")
```

The result is:

``` python
Traceback (most recent call last):
  File "/usr/local/lib/python2.7/site-packages/tablib/formats/_csv.py", line 48, in import_set
    dset.append(row)
  File "/usr/local/lib/python2.7/site-packages/tablib/core.py", line 668, in append
    self.rpush(row, tags)
  File "/usr/local/lib/python2.7/site-packages/tablib/core.py", line 652, in rpush
    self.insert(self.height, row=row, tags=tags)
  File "/usr/local/lib/python2.7/site-packages/tablib/core.py", line 643, in insert
    self._validate(row)
  File "/usr/local/lib/python2.7/site-packages/tablib/core.py", line 283, in _validate
    raise InvalidDimensions
tablib.core.InvalidDimensions
```

It may not be a good idea to actually represent ragged data in a Dataset, but it should be possible to import ragged CSV text into a uniform-width Dataset by filling in empty elements to make all rows match the width of the widest row.

klothe:
See #227

kennethreitz:
I'm torn about this. We could make it possible to provide `None' as a value for extra values, but I'd like to think that this type of data simply doesn't belong with this library.

klothe:
I've been using this feature to import rectangular data sets from CSV files where separator characters for trailing blank cells are omitted. In particular, [Tabula](https://github.com/tabulapdf/tabula) outputs CSV files like this when extracting tables from PDFs.

I agree that Tablib shouldn't be used for working with ragged data sets, but these are more like rectangular data sets with variant formatting (because columns are meaningful), so filling in extra blank cells is the right behavior. I found it easier to handle that inside Tabilb than to pre-process the files before importing them to get around the InvalidDimensions error.

dcfix:
Isn't a ragged CSV file really just a rectangular data set with some primitive compression applied?

brihuega:
This issue still raises when the first row is shorter.

`import tablib
ds = tablib.import_set("a,b\nx,y,z", format='csv', headers=False)`

`Traceback (most recent call last):
[...]

PR #647, opened 2026-06-23
Normalize ragged CSV to the widest row, not the header width

binggao1230:
## Summary

CSV import is documented (issue #226) to normalize ragged rows "to the width of the widest row by filling in empty elements". In practice it only pads rows up to the width of the **first/header** row, so a data row *wider* than the header raises `InvalidDimensions`:

```python
>>> import tablib
>>> tablib.import_set("a,b\nc,d,e\n", format="csv")
tablib.exceptions.InvalidDimensions
>>> tablib.import_set("a,b\nc,d,e\n", format="csv", headers=False)
tablib.exceptions.InvalidDimensions
```

The analogous *shorter* case is handled — short rows are padded — so the behaviour is inconsistent, and the existing `test_csv_import_set_ragged` only exercises the header-is-widest direction.

## Fix

Read the header and data rows, compute the width as the maximum across **all** rows (header included), then pad the header and every data row to that width. Empty data lines are still skipped, and an empty/absent header still yields `headers=None`.

```python
>>> dataset = tablib.import_set("a,b\nc,d,e\n", format="csv")
>>> dataset.headers
['a', 'b', '']
>>> list(dataset[0])
['c', 'd', 'e']
```

## Verification

- Reproduced on `master` (`64f0201`): a wider-than-header row raises `InvalidDimensions`; with the fix it normalizes to the widest width (with and without headers).
- Added `test_csv_import_set_ragged_wider_than_header` and `test_csv_import_set_ragged_wider_no_headers`; both fail before the fix and pass after.
- Full suite: **176 passed** (174 baseline + the 2 new tests), no regressions. `ruff check` clean.
- Existing behaviour preserved: shorter-row padding, the empty-line-skipping `test_csv_import_set_ragged`, and empty-stream (`headers=None`) all unchanged.

This completes the ragged-CSV feature from #226 for the wider-row case.

---

This pull request was prepared with the assistance of AI, under my direction and review.

### DIFF
--- a/src/tablib/formats/_csv.py
+++ b/src/tablib/formats/_csv.py
@@ -43,15 +43,28 @@ class CSVFormat:
         kwargs.setdefault('delimiter', cls.DEFAULT_DELIMITER)
 
         rows = csv.reader(in_stream, **kwargs)
+
+        header_row = None
+        data_rows = []
         for i, row in enumerate(rows):
             if i < skip_lines:
                 continue
             if i == skip_lines and headers:
-                dset.headers = row
+                header_row = row
             elif row:
-                if i > 0 and len(row) < dset.width:
-                    row += [''] * (dset.width - len(row))
-                dset.append(row)
+                data_rows.append(row)
+
+        # Ragged rows are normalized to the width of the widest row by filling
+        # in empty elements, not just to the width of the header row (issue #226).
+        width = max((len(r) for r in data_rows), default=0)
+        if header_row is not None:
+            width = max(width, len(header_row))
+            dset.headers = header_row + [''] * (width - len(header_row))
+
+        for row in data_rows:
+            if len(row) < width:
+                row = row + [''] * (width - len(row))
+            dset.append(row)
 
     @classmethod
     def detect(cls, stream, delimiter=None):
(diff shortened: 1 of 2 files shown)


MohammedAlkindi:
Ran this on Windows 11 (Python 3.13.13), since it hadn't had a review yet.

At master `c8ff4e1`, `tablib.import_set('a,b\nc,d,e\n', format='csv')` raises `InvalidDimensions` from `core.py:278`, exactly as described. On this branch (`3d3617a`, merge-base `64f0201`) the same call returns headers `['a', 'b', '']` and row `('c', 'd', 'e')`.

Both new tests fail on master with `InvalidDimensions` (checked by copying `tests/test_tablib.py` onto master) and pass here. Full suite on this branch: 175 passed, 1 failed — the failure is the pre-existing timezone-dependent `test_getDate_datetime_timestamp`, which fails identically on master on a UTC+4 machine (see #675), so no new failures. `ruff check` is clean on `_csv.py`.

Edge behavior matches the PR text: empty lines are still skipped, an empty stream still gives `headers=None`, and shorter-than-header rows still pad. Two observations for the record: input with CRLF line endings behaves identically to LF, and the fix also applies to TSV imports since `TSVFormat` subclasses `CSVFormat` — may be worth a word in the description if that's intended.

PR #649, opened 2026-07-03
Raise TypeError for invalid Dataset instance

koteshyelamati:
## Problem

`stack()` and `stack_cols()` silently return `None` when passed a non-`Dataset` argument:

```python
ds = tablib.Dataset()
result = ds.stack("not a dataset")
print(result)  # None — caller has no idea the operation failed
```

This makes bugs very hard to diagnose: callers receive `None` without any indication that the argument was invalid.

## Fix

Raise `TypeError` with a clear message instead of silently returning `None`.

```python
ds = tablib.Dataset()
ds.stack("not a dataset")
# TypeError: 'other' must be a Dataset instance
```

### DIFF
--- a/src/tablib/core.py
+++ b/src/tablib/core.py
@@ -746,7 +746,7 @@ class Dataset:
         ``Dataset`` instance."""
 
         if not isinstance(other, Dataset):
-            return
+            raise TypeError("'other' must be a Dataset instance")
 
         if self.width != other.width:
             raise InvalidDimensions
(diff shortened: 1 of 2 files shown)


claudep (changes requested):
Thanks, but this is missing a regression test.

koteshyelamati:
Thanks @claudep — added test_stack_rejects_non_dataset covering both stack() and stack_cols() with a non-Dataset argument. Ran the stack test group locally with pytest: 3 passed.

claudep (approved):
Thanks, I think it is in good shape now.

PR #658, opened 2026-07-15
Only strip the leading '=' when escaping XLSX formulae

vidigoat:
### Summary

When exporting XLSX with `escape=True`, formula cells have **every** `=` removed:

```python
cell.value = cell.value.replace("=", "")
```

but the option is documented to remove only the **leading** `=` (`_xlsx.py` docstring: *"formulae will have the leading '=' character removed"*, and the original feature in #540). A legitimate comparison formula like `=A1=B1` is therefore corrupted to `A1B1` instead of `A1=B1`.

Removing just the leading `=` already neutralises the formula (the cell is no longer treated as a formula by Excel), so stripping the inner `=` is unnecessary data loss.

### Fix

Replace `cell.value.replace("=", "")` with `cell.value[1:]`. The leading `=` is guaranteed present by the existing `startswith('=')` guard, so this only drops the leading character. The existing `=SUM(1+1)` → `SUM(1+1)` behaviour is unchanged.

### Test

Added `test_xlsx_export_set_escape_formulae_keeps_inner_equals`: `=A1=B1` with `escape=True` now yields `A1=B1`. It fails before the change (`'A1=B1' != 'A1B1'`) and passes after. All existing escape/XLSX tests still pass. Added myself to `AUTHORS`.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -38,4 +38,5 @@ Here is a list of past and present much-appreciated contributors:
     Tommy Anthony
     Tsuyoshi Hombashi
     Tushar Makkar
+    Vidit Patankar
     Yunis Yilmaz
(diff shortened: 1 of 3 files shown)


claudep (approved):
Looks good, thanks!

PR #659, opened 2026-07-17
fix: IndexError exporting a zero-width dataset to reStructuredText

santhreal:
Exporting a zero-width dataset (rows present, no columns) to rst raises `IndexError`. `export_set` guards the empty case with `if not dataset.dict`, but a zero-width dataset has `dict == [[]]` which is truthy, so it falls through to `get_col(0)`/`column_widths[0]` and indexes an empty row.

Such a dataset is valid: `[[]]` is valid JSON that imports cleanly (`import_set` sets `.dict` directly), and every other text format degrades gracefully (csv/tsv give a blank line, json gives `[[]]`, html/latex give an empty table). Only rst crashes. This also returns `''` when the dataset has no columns.

```python
import tablib
tablib.import_set('[[]]', format='json').export('rst')  # IndexError before, '' after
```

### DIFF
--- a/src/tablib/formats/_rst.py
+++ b/src/tablib/formats/_rst.py
@@ -241,7 +241,7 @@ class ReSTFormat:
         True
 
         """
-        if not dataset.dict:
+        if not dataset.dict or not dataset.width:
             return ''
         force_grid = kwargs.get('force_grid', False)
         max_table_width = kwargs.get('max_table_width', cls.MAX_TABLE_WIDTH)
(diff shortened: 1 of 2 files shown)


hugovk:
You've opened many PRs across these repos two weeks ago, and then closed them with no comment:

https://github.com/python-humanize/humanize/pull/352
https://github.com/python-humanize/humanize/pull/353
https://github.com/prettytable/prettytable/pull/483
https://github.com/prettytable/prettytable/pull/484
https://github.com/jazzband/tablib/pull/659
https://github.com/jazzband/tablib/pull/660
https://github.com/jazzband/tablib/pull/661

Please could you explain why?

Also hundreds in other repos:

image

PR #665, opened 2026-07-20
Convert XLS cell values when importing a Databook

vineethsaivs:
## The bug

`XLSFormat.import_set` reads cells through a `cell_value` helper that turns Excel date serials into `datetime` objects and error codes into strings like `#N/A`. `XLSFormat.import_book` re-implemented the row loop with raw `sheet.row_values(i)` and never got that helper, so the same file returns different values depending on whether you load it as a `Dataset` or a `Databook`.

Using the fixtures already in the test suite:

```
dates.xls    'birth_date'
   Dataset  -> datetime.datetime(2015, 4, 12, 0, 0)
   Databook -> 42106.0

errors.xls   'not available (formula)'
   Dataset  -> '#N/A'
   Databook -> 42
```

That `42` is the symptom from #202 ("Tablib import Excel NA() values as 42"), which was fixed on the `import_set` path only; it still reproduces today through `Databook`.

XLS is the only spreadsheet format where the two paths differ. Both XLSX and ODS delegate `import_set` and `import_book` to a shared `import_sheet`, so they cannot drift. The existing `test_xls_date_import` and `test_xls_import_with_errors` assert the converted values, but only via `Dataset`, and the one XLS Databook test only checks a sheet title, which is why this went unnoticed.

## The fix

Extract an `import_sheet` classmethod and call it from both entry points, mirroring the structure `_xlsx.py` and `_ods.py` already use. The cell conversion becomes a `read_cell` classmethod, matching the `read_cell` in `_ods.py`, with the workbook `datemode` passed in explicitly rather than captured from an enclosing scope (`sheet.book` is not reliable under xlrd's `on_demand`).

Header handling is deliberately unchanged. `import_set` reads headers with raw `row_values` today and the shared helper keeps doing that, so `Databook` now matches `Dataset` exactly rather than changing both.

## Tests

Two regression tests load the existing `dates.xls` and `errors.xls` fixtures as a `Databook` and assert the same values the `Dataset` tests already assert. Both fail before this change (`42106.0 != datetime.datetime(2015, 4, 12, 0, 0)`) and pass after.

```
$ pytest tests/ -q
182 passed
```

Full suite passes before and after; the count goes from 180 to 182 with the two new tests. Added myself to `AUTHORS` per the contributing guide.

### DIFF
--- a/AUTHORS
+++ b/AUTHORS
@@ -39,4 +39,5 @@ Here is a list of past and present much-appreciated contributors:
     Tsuyoshi Hombashi
     Tushar Makkar
     Vidit Patankar
+    Vineeth Sai
     Yunis Yilmaz
--- a/src/tablib/formats/_xls.py
+++ b/src/tablib/formats/_xls.py
@@ -93,22 +93,19 @@ class XLSFormat:
         return stream.getvalue()
 
     @classmethod
-    def import_set(cls, dset, in_stream, headers=True, skip_lines=0):
-        """Returns databook from XLS stream."""
+    def read_cell(cls, value, type_, datemode):
+        """Returns the Python value of a cell, based on its xlrd type."""
+        if type_ == xlrd.XL_CELL_ERROR:
+            return xlrd.error_text_from_code[value]
+        elif type_ == xlrd.XL_CELL_DATE:
+            return xldate_as_datetime(value, datemode)
+        return value
 
-        dset.wipe()
-
-        xls_book = xlrd.open_workbook(file_contents=in_stream.read())
-        sheet = xls_book.sheet_by_index(0)
+    @classmethod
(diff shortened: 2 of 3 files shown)


vineethsaivs:
The only red check here is `label`: *"Label error. Requires at least 1 of: changelog: Added, changelog: Changed, ..."*. I do not have rights to label the PR myself, so could a maintainer add `changelog: Fixed`? Everything else (tests, codecov) is green.

Recap: `Databook` import goes through the sheet rows directly, so XLS cell values are never passed through the same conversion the single-`Dataset` path applies. Dates, times and booleans therefore come back as raw floats/ints in a `Databook` but as proper Python objects in a `Dataset` from the same file. The fix routes both paths through the same conversion, with a regression test covering the mismatch.

claudep (approved):
Looks good, thanks for this fix!

PR #672, opened 2026-08-07
Add .gitattributes for cross-platform line-ending normalization

chernojagne:
Add a `.gitattributes` file to ensure consistent line-ending handling across platforms.

**Why:** tablib's CI matrix runs tests on Ubuntu, macOS, and Windows. Without `* text=auto`, line-ending normalization is git's built-in default but platform-dependent — contributors on Windows may commit CRLF-normalized files that break diffs or cause spurious test failures. The file also marks binary assets (images, fonts, archives, lockfiles, Python bytecode) explicitly so git doesn't attempt text diff or normalization on them.

**What changed:** A single new `.gitattributes` file using the `standard` profile:
- `* text=auto` — universal line-ending normalization
- Binary markers for images, fonts, archives, compiled objects, and Python-ecosystem lockfiles (`uv.lock`, `Pipfile.lock`, `poetry.lock`)

No existing files are modified.

---

This change was generated by [Autar](https://autar.ai) and reviewed by a human before submission.


webknjaz:
I clicked on
> generated by [Autar](https://autar.ai)

and it literally says the on the front page 

> Don’t maintain it

🤯🤯🤯😭😭😭😭😭 c'mon...

chernojagne:
Fair point, and thank you for calling it out directly. The tagline is about not being burdened by your *own* project's maintenance — but the irony of creating review work for *other* maintainers in the process is real and not lost on me.

If you'd rather not deal with this PR, just say so and I'll close it. Happy to defer entirely to your preference here.

webknjaz:
It's just harmful to send out mass submissions w/o making sure you understand each project's needs and are a part of their active contributor base.

PR #673, opened 2026-08-14
Raise UnsupportedFormat for a list with non-uniform row types in _set_dict

eeshsaxena:
Hit this loading some YAML that came back from an external source. A list whose first item is a mapping but which contains a later non-mapping item, e.g.

    - {id: 0, name: n0}
    - {id: 1, name: n1}
    -

(that trailing bare `-` parses to `None`) raises `AttributeError: 'NoneType' object has no attribute 'values'` from `Dataset._set_dict` instead of a tablib error.

`_set_dict` decides the shape from `pickle[0]` alone and then assumes every row has that type: the dict branch does `list(row.values())` on each row, and the list-of-rows branch does `Row(row)`, so a later element of a different type raises a bare `AttributeError` (or `TypeError` for the list branch) rather than the `UnsupportedFormat` that malformed input normally gets. I added a per-row type check to both branches.

Added a test next to the existing YAML import tests; it fails on master with the AttributeError and passes with the change, and the dict/yaml/json/csv tests still pass. Found it while fuzzing the format importers with mutated input.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise UnsupportedFormat for a list with non-uniform row types in _set_dict`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining tablib!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

hugovk:
There's a cost to maintainers in adding to the backlog, even if we didn't get a chance to review this (or https://github.com/jazzband/tablib/pull/674 or https://github.com/jazzband/tablib/pull/678).

May I suggest refraining on opening future PRs if that not that important for you that they can't wait three weeks.

eeshsaxena:
@hugovk I am extremely sorry that you took offence but a ping by maintainer is customary for a pr that is 3 weeks old which is light years in technological words 
Once again I am sorry

hugovk:
I'm not talking about the nudge, but that's also unnecessary. Light years are distance not time, but if you mean three weeks is an extremely long time to go without review, that's simply not the case in many open source projects. And the recent LLM-enabled surge isn't helping.

PR #674, opened 2026-08-14
Raise UnsupportedFormat for a malformed DBF stream

eeshsaxena:
Loading a malformed or truncated DBF file surfaces low-level errors straight out of `Dataset().load(..., format='dbf')` rather than tablib's own `UnsupportedFormat`. Depending on how the bytes are broken you get a `struct.error`, `IndexError`, `UnicodeDecodeError`, `KeyError`, etc. from the vendored dbfpy parser.

`DBFFormat.detect` already wraps the parser in `try/except Exception: return False`, so tablib already treats any DBF parsing error as "not a valid DBF". I made `import_set` consistent: it wraps the parse and raises `UnsupportedFormat` on any parser error. I also build the rows before touching the dataset so a failure part-way through doesn't leave a half-populated Dataset.

Added a test (truncated header + non-DBF bytes) that raises the raw IndexError on master and passes with the change; valid DBF import/export and the rest of the DBF tests are unchanged. Found it by fuzzing the format importers.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise UnsupportedFormat for a malformed DBF stream`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining tablib!

eeshsaxena:
Closing this to clear out my older open PRs. Nothing wrong with the change on my side, I am just tidying up a backlog. If it is still useful to you, say so and I will reopen it and rebase it on current main.

PR #678, opened 2026-08-17
Raise UnsupportedFormat for a non-dict row in a list of records

eeshsaxena:
When the `dict` setter (used by JSON and YAML import) gets a list whose first item is a dict but which also contains a non-dict item, it calls `row.values()` on that item and raises a bare `AttributeError`. A YAML document like `- a: 1`/`- null` or JSON `[{"a": 1}, null]` triggers it. I check each row is a dict and raise `UnsupportedFormat` instead, matching what the setter already raises for other malformed input.


eeshsaxena:
Hi! Gentle nudge on this one whenever you have some bandwidth. It's a small, self-contained fix (`Raise UnsupportedFormat for a non-dict row in a list of records`), and it's currently mergeable with no conflicts. No urgency at all, and I'm happy to make any changes you'd like. Thanks for maintaining tablib!

Issue #199, opened 2015-09-01
Dataset.transpose() fails if there are repeated headers


rhunwicks:
If the same value appears multiple times in the headers of a Dataset, then when you `transpose()` it the values from the first time that header occurs are copied to all instances of it in the new Dataset. Although normally we wouldn't expect to have repeated headers, in the case of `transpose()` it seems quite likely, given that the initial header row is actually the first column of the Dataset you want. 

I would expect `Dataset.transpose().transpose()` to give me the original Dataset back. However:

```python
import tablib
data = tablib.Dataset()
data.headers = ("first_name", "last_name", "age")
data.append(('John', 'Adams', 90))
data.append(('George', 'Washington', 67))
data.append(('John', 'Tyler', 71))
data.transpose().transpose().dict == data.dict
```
Which gives:
```
False
```
If we investigate in more detail:
```
data.transpose().transpose()['first_name']
data.transpose().transpose()['last_name']
data.transpose().transpose()['age']
```
We get:
```
['John', 'George', 'John']
['Adams', 'Washington', 'Adams']
[90, 67, 90]
```

The broken bit of `transpose()` is:

```python
        for column in self.headers:

            if column == self.headers[0]:
                # It's in the headers, so skip it
                continue

            # Adding the column name as now they're a regular column
            row_data = [column] + self[column]
            row_data = Row(row_data)
            _dset.append(row=row_data)
```

If you replace that with:

```python
        for index, column in enumerate(self.headers):

            if column == self.headers[0]:
                # It's in the headers, so skip it
                continue

            # Adding the column name as now they're a regular column
            # Use `get_col(index)` in case there are repeated values
            row_data = [column] + self.get_col(index)
            row_data = Row(row_data)
            _dset.append(row=row_data)
```

then it works as expected.

kennethreitz:
Thank you very much! This has now been integrated.

kennethreitz:
:sparkles: :cake: :sparkles:

punchagan:
I found a related bug to this, though this kind of data may not occur frequently. The bug was found using @HypothesisWorks.

```python
data = tablib.Dataset(('fname', 'Doe', '10'), headers=['fname', 'lname', 'age'])
print(data, data.transpose().transpose(), sep='\n\n')
```

```markdown
fname|lname|age
-----|-----|---
fname|Doe  |10 

fname|lname|age
-----|-----|---
```

PR #682, opened 2026-09-07
Preserve columns matching the first header during transpose

Tamako-C:
`Dataset.transpose()` currently skips every column whose header matches the first header. Only the first column belongs in the new header row; later columns with the same name must still become data rows.

Skip the first column by position. Add regression tests for a repeated first header and for the case where an original cell value matches the first header, causing data loss after two transposes.

This addresses the remaining case reported by punchagan in https://github.com/jazzband/tablib/issues/199#issuecomment-250525292. The earlier PR #199 is already merged; this is a follow-up fix.

Validation:

- Both regression tests fail against the original source and pass after the fix.
- Full suite on Python 3.12.14 with pandas 3.0.1 and all required format dependencies: 185 passed with process-local `TZ=UTC`, without skipped tests.
- In the host's UTC+8 timezone: 184 passed and one existing date-conversion test failed. Running that same test against the original HEAD reproduces the failure; it matches https://github.com/jazzband/tablib/pull/675. No system timezone or test expectations were changed.
- Changed Python files pass local Ruff 0.11.9. After submission, [pre-commit.ci passed](https://results.pre-commit.ci/run/github/1535009/1788801161.47FxFT6dSW-bDCdHoFZjsg) using the repository's configuration, which pins Ruff 0.15.20.
- `git diff --check` passes.

AI assistance: OpenAI Codex assisted with the implementation, regression tests, and this PR description.


Voyagerroc-Lab (approved):
Skipping the first header by position preserves later columns when header names repeat, and the round-trip regression covers cell values equal to the first header. The author reports both regressions fail on the original source and pass with the change; the full suite passed 185 tests in UTC. Current pre-commit and documentation checks pass. Review performed with OpenAI Codex at the account owners request.

PR #686, opened 2026-09-20
Optimize Dataset.subset() lookups

### LINKED ISSUE #685: Optimize Dataset.subset() to avoid repeated lookups
`Dataset.subset()` becomes slow on larger datasets because it repeatedly looks up the same column positions for every row. It also searches a list each time it checks whether a row was selected.

These repeated searches are unnecessary. We can calculate the column positions once and store the selected row numbers in a set for faster lookup.

This reduces the time complexity from approximately O(n³) to O(n²) while keeping the existing behavior unchanged.

harshitgupta31415:
## What changed?

`Dataset.subset()` was repeatedly looking up the same rows and columns while creating a subset. This made it noticeably slower for larger datasets.

This change:

- remembers the selected rows for quicker checks
- finds each selected column's position only once
- processes only the rows that are needed

For a 400×400 dataset, the execution time improved from about 1.03 seconds to 0.03 seconds.

A test was added to make sure the existing selection behavior remains unchanged. All 184 tests pass.

Fixes #685


hugovk:
Please comment with a benchmark for this.

harshitgupta31415:
Thanks, @hugovk! I benchmarked the base commit (`a36c965`) against this PR (`98b9bd1`) using Python 3.14.5 on Windows. Dataset construction was completed before timing, so the results measure only `Dataset.subset()`.

### Square datasets

Each result is the median of five runs.

| Dataset size | Base | This PR | Speedup |
| --- | ---: | ---: | ---: |
| 100×100 | 0.017916 s | 0.002422 s | 7.4× |
| 200×200 | 0.116337 s | 0.010250 s | 11.3× |
| 400×400 | 0.973558 s | 0.024506 s | 39.7× |

### Rectangular dataset

I also measured two rectangular datasets using the median of ten runs:

| Dataset size | Base | This PR | Speedup |
| --- | ---: | ---: | ---: |
| 4,000×50 | 0.235769 s | 0.026538 s | 8.9× |
| 400×4,092 | 106.261264 s | 0.344110 s | 308.8× |

For an n×n selection, the change reduces runtime from approximately O(n³) to O(n²). It uses a row set and a precomputed column-index list, but auxiliary space remains O(n); the O(n²) returned dataset still dominates total memory usage.

<details>
<summary>Benchmark code</summary>

```python
import platform
from statistics import median
from timeit import repeat

from tablib import Dataset


print(f"Python {platform.python_version()}")

cases = (
    (100, 100, 5),
    (200, 200, 5),
    (400, 400, 5),
    (4_000, 50, 10),
    (400, 4_092, 10),
)

for row_count, column_count, repetitions in cases:
    headers = [f"column_{index}" for index in range(column_count)]
    row = tuple(range(column_count))
    dataset = Dataset(*(row for _ in range(row_count)), headers=headers)

    elapsed = median(
        repeat(
            lambda: dataset.subset(
                rows=range(row_count),
                cols=headers,
            ),
            repeat=repetitions,
            number=1,
        )
    )

    print(f"{row_count} rows x {column_count} columns: {elapsed:.6f}s")
```

</details>

PR benchmark

Base benchmark

Base benchmark

hugovk:
@harshitgupta31415 Please fix the markup in your last comment.

harshitgupta31415:
Thanks for pointing that out, @hugovk. I've fixed the markup in the benchmark comment, and the tables, code block, and screenshots should now render correctly.

hugovk:
Can you check the behaviour change when `cols=[]` or every requested column is missing?

harshitgupta31415:
Good catch, @hugovk  this was changing the behavior. I’ve fixed it so both cols=[] and an all-missing column list behave the same as old again. Added tests for both cases too.

